PR #3, opened 2022-04-04
Add mysqlsh wrapper

paulomach:
# Issue
Necessity to have common function to wrap mysqlsh calls

# Solution
Inherited working code from reactive machine charm (openstack team).

# Context
MySQL setup/configuration calls are execute using mysql-shell cli.
There's still no tests, since we need it to work at the same time on different parts of the code.

# Release Notes
* Imported code from [openstack machine charm](https://github.com/openstack/charm-mysql-innodb-cluster/blob/d8d1dc0527046cc3dd41f4394575f14e9675f230/src/lib/charm/openstack/mysql_innodb_cluster.py#L1672)
* refactor to add type hints
* refactor to current standard docstrings

### DIFF
--- a/src/mysqlsh_helpers.py
+++ b/src/mysqlsh_helpers.py
@@ -27,7 +27,8 @@ class MySQL:
     def mysqlsh_bin(self) -> str:
         """Determine binary path for MySQL Shell.
 
-        :returns: Path to binary mysqlsh
+        Returns:
+            Path to binary mysqlsh
         """
         # Allow for various versions of the mysql-shell snap
         # When we get the alias use /snap/bin/mysqlsh
@@ -42,7 +43,8 @@ class MySQL:
     def mysqlsh_common_dir(self) -> str:
         """Determine snap common dir for mysqlsh.
 
-        :returns: Path to common dir
+        Returns:
+            Path to common dir
         """
         return "/root/snap/mysql-shell/common"
 
@@ -69,11 +71,13 @@ class MySQL:
     def run_mysqlsh_script(self, script: str) -> AnyStr:
         """Execute a MySQL shell script.
 
-        :param script: Mysqlsh script
-        :raises subprocess.CalledProcessError: Raises CalledProcessError if the
-                                               script gets a non-zero return
-                                               code.
-        :returns: subprocess output
+        Raises CalledProcessError if the script gets a non-zero return code.
+
+        Args:
+            script: Mysqlsh script string
+
+        Returns:
+            Byte string subprocess output
         """
         if not os.path.exists(self.mysqlsh_common_dir):
             # Pre-execute mysqlsh to create self.mysqlsh_common_dir
--- a/src/mysqlsh_helpers.py
+++ b/src/mysqlsh_helpers.py
@@ -40,7 +40,7 @@ class MySQL:
 
     @property
     def mysqlsh_common_dir(self):
-        """Determine snap common dir for mysqlsh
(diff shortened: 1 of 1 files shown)


src/mysqlsh_helpers.py
```diff
@@ -38,3 +66,38 @@ def create_cluster(self):
     def add_instance_to_cluster(self):
         """Add an instance to the InnoDB cluster."""
         pass
+
+    def run_mysqlsh_script(self, script, mode="python"):
```
shayancanonical, line 70: should this be a private method?

src/mysqlsh_helpers.py
```diff
@@ -38,3 +66,38 @@ def create_cluster(self):
...
+            # If we don't do this the real execution will fail with an
+            # ambiguous error message. This will only ever execute once.
+            cmd = [self.mysqlsh_bin, "--help"]
+            subprocess.check_call(cmd, stderr=subprocess.PIPE)
+
+
+        # Use the self.mysqlsh_common_dir dir for the confined
+        # mysql-shell snap.
+        with tempfile.NamedTemporaryFile(
+            mode="w", suffix=".py", dir=self.mysqlsh_common_dir
```
shayancanonical, line 94: should there be a suffix of `.py`? it may be a bit misleading

src/mysqlsh_helpers.py
```diff
@@ -19,6 +22,31 @@ class MySQL:
...
+        return "/snap/bin/mysql-shell"
+
+    @property
+    def mysqlsh_common_dir(self):
+        """Determine snap common dir for mysqlsh
+
+        :returns: Path to common dir
+        :rtype: str
+        """
+        return "/root/snap/mysql-shell/common"
```
shayancanonical, line 48: will we always run `mysqlsh` as `root`? it is downloaded as `root`, but i wonder if we should do things with user `mysql` in a future PR

src/mysqlsh_helpers.py
```diff
@@ -38,3 +66,38 @@ def create_cluster(self):
...
+
+    def run_mysqlsh_script(self, script, mode="python"):
+        """Execute a MySQL shell script
+
+        :param script: Mysqlsh script
+        :type script: str
+        :param mode: Script type (python, or sql)
+        :type mode: str
+        :side effect: Calls subprocess.check_output
+        :raises subprocess.CalledProcessError: Raises CalledProcessError if the
```
shayancanonical, line 78: to confirm: this is actually the docstring format on the other database operators and actually something we want to adopt

src/mysqlsh_helpers.py
```diff
@@ -38,3 +65,30 @@ def create_cluster(self):
...
+
+    def run_mysqlsh_script(self, script: str) -> AnyStr:
+        """Execute a MySQL shell script.
+
+        :param script: Mysqlsh script
+        :raises subprocess.CalledProcessError: Raises CalledProcessError if the
+                                               script gets a non-zero return
+                                               code.
+        :returns: subprocess output
+        """
```
marceloneppel, line 77: As Shayan commment, this can be changed to something like [this](https://github.com/canonical/mongodb-operator/blob/ac6c0ef8b2cb24f607b26cffa1ea0e94d5748692/src/mongod_helpers.py#L302) to have a pattern in the docstrings.

marceloneppel (commented):
LGTM. One comment about the docstrings.

marceloneppel (approved):

shayancanonical (approved):

PR #4, opened 2022-04-05
DPE-145 Implement a helper method to configure instance for InnoDB

shayancanonical:
Part of [Jira ticket](https://warthogs.atlassian.net/jira/software/c/projects/DPE/boards/390?modal=detail&selectedIssue=DPE-145)

# Issue
We need to implement `configure_instance` for the MySQL helper class.

# Solution
- Utilize the `_run_mysqlsh_script` helper method to configure the MySQL instance for InnoDB
- Implement a `_wait_until_mysql_connection` helper method to wait till a MySQL connection is obtained (to test connection to DB after restarting as part of instance configuration)
- Add unit tests

# Context
We are using `mysqlsh` in python mode to configure the instance, and then restarting it by running the `RESTART` mysql command

# Release Notes
- Implement `configure_instance` method for the MySQL helper class


src/mysqlsh_helpers.py
```diff
@@ -56,19 +73,45 @@ def configure_mysql_users(self):
         """
         pass
 
-    def configure_instance(self):
+    def configure_instance(self) -> None:
         """Configure the instance to be used in an InnoDB cluster."""
-        pass
+        _script = f"""
+dba.configure_instance('{self.cluster_admin_user}:{self.cluster_admin_password}@{self.instance_address}')
```
paulomach, line 79: As mentioned, assemble script as tuple, then join it with:

```python
"\n".join(_script)
```

src/mysqlsh_helpers.py
```diff
@@ -56,19 +73,45 @@ def configure_mysql_users(self):
...
 
-    def run_mysqlsh_script(self, script: str) -> AnyStr:
+    @retry(reraise=True, stop=stop_after_delay(30), wait=wait_fixed(5))
+    def _wait_until_mysql_connection(self) -> None:
+        """Wait until a connection to MySQL has been obtained.
+
+        Retry every 5 seconds for 30 seconds if there is an issue obtaining a connection.
+        """
+        _script = f"""
+my_shell = shell.connect('{self.cluster_admin_user}:{self.cluster_admin_password}@{self.instance_address}')
```
paulomach, line 109: It's not possible to set a timeout to `shell.connect` ?

paulomach (approved):
Minor comments. Tests are mint.
LGTM

src/mysqlsh_helpers.py
```diff
@@ -56,19 +73,45 @@ def configure_mysql_users(self):
...
         pass
 
-    def add_instance_to_cluster(self):
+    def add_instance_to_cluster(self) -> None:
         """Add an instance to the InnoDB cluster."""
         pass
 
-    def run_mysqlsh_script(self, script: str) -> AnyStr:
+    @retry(reraise=True, stop=stop_after_delay(30), wait=wait_fixed(5))
+    def _wait_until_mysql_connection(self) -> None:
```
marceloneppel, line 103: Shouldn't the return type be `AnyStr`? Otherwise, you can remove the return statement at the end of this function.

marceloneppel (approved):
LGTM! One minor comment.

tests/unit/test_mysqlsh_helpers.py
```diff
@@ -0,0 +1,101 @@
...
+        self.mysql = MySQL(
+            root_password, cluster_admin_user, cluster_admin_password, instance_address
+        )
+
+    @patch("os.path.exists", return_value=False)
+    @patch("mysqlsh_helpers.MySQL.mysqlsh_bin", new_callable=PropertyMock)
+    @patch("subprocess.check_call")
+    @patch("tempfile.NamedTemporaryFile")
+    @patch("subprocess.check_output")
+    def test_run_mysqlsh_script(
```
zmraul, line 27: Hm, is anything really being tested here? All the calls are mocked so nothing real is happening. No suggestions though, I understand that this is needed to get coverage up, but sometimes it feels unnecessary.

zmraul (approved):
Looks good! I would add the suggestions from Paulo and Marcelo

src/mysqlsh_helpers.py
```diff
@@ -56,19 +73,45 @@ def configure_mysql_users(self):
...
+        """Wait until a connection to MySQL has been obtained.
+
+        Retry every 5 seconds for 30 seconds if there is an issue obtaining a connection.
```
delgod, line 106: if MySQL is doing crash recovery of big DB during mysqld start -  it can take days

shayancanonical: if that's the case, wouldn't it be a good idea to raise the exception and move the unit into `BlockedStatus`?

src/mysqlsh_helpers.py
shayancanonical: there's a default connect timeout of 10s. if the server is unavailable, i believe that the command will fail right away - and result in retries every 5s for 30s. however, if the server is overwhelmed, the command will timeout in 10s and then we will still retry every 5s - but only until the aggregate time is less than 30s (at least that is how I think it would work)

src/mysqlsh_helpers.py
shayancanonical: addressed in 46822799e0a54478e31361bb4cbea6fd70c4e454

src/mysqlsh_helpers.py
shayancanonical: good point, removed return type in 46822799e0a54478e31361bb4cbea6fd70c4e454. we will add it back if we find a need for it in the future

tests/unit/test_mysqlsh_helpers.py
shayancanonical: good call. removed the superfluous tests in 46822799e0a54478e31361bb4cbea6fd70c4e454

PR #5, opened 2022-04-05
Dpe 144 user creation methods

paulomach:
# Issue

Jira link is (DPE-144](https://warthogs.atlassian.net/browse/DPE-144).

For mysql, we need to create a "clusteradmin" user with full privileges but that can be run only from the instance, playing the role of the "operator".

# Solution

# Context

For mysql, we need to create a `clusteradmin` user with full privileges but that can be run only from the instance, playing the role of the "operator".
We also need to create the general `root` user for external access (i.e. mysql-router, other apps). This user must not have the privilege to allow it change other users passwords, hence protecting access lost for the cluster.
Additionally, we must set password for all of these users.

# Testing

# Release Notes

### DIFF
--- a/src/mysqlsh_helpers.py
+++ b/src/mysqlsh_helpers.py
@@ -26,18 +26,25 @@ class MySQL:
     creation and configuration of MySQL InnoDB clusters via Group Replication.
     """
 
-    def __init__(self, root_password: str, cluster_admin_user: str, cluster_admin_password: str):
+    def __init__(
+        self,
+        root_password: str,
+        cluster_admin_user: str,
+        cluster_admin_password: str,
+        instance_address: str,
+    ):
         """Initialize the MySQL class.
 
         Args:
             root_password: Password for the 'root' user
             cluster_admin_user: User name for the cluster admin user
             cluster_admin_password: Password for the cluster admin user
+            instance_address: Address of the targeted instance
         """
         self.root_password = root_password
         self.cluster_admin_user = cluster_admin_user
         self.cluster_admin_password = cluster_admin_password
-        self.instance_address = None
+        self.instance_address = instance_address
 
     @property
     def mysqlsh_bin(self) -> str:
@@ -72,21 +79,27 @@ class MySQL:
     def configure_mysql_users(self):
         """Configure the MySQL users for the instance.
 
-        Creates a 'clusteradmin' user with the appropriate privileges and
-        revokes certain privileges from the 'root' user.
+        Creates base `root@%` and `clusteradmin@instance_address` user with the
+        appropriate privileges, and reconfigure `root@localhost` user password.
         """
         _script = (
-            f'shell.connect("root:{self.root_password}@localhost")',
-            f"dba.session.run_sql(\"CREATE USER '{self.cluster_admin_user}'@'%' IDENTIFIED BY '{self.cluster_admin_password}' ;\")",
-            f"dba.session.run_sql(\"GRANT ALL ON *.* TO '{self.cluster_admin_user}'@'%' WITH GRANT OPTION ;\")",
-            'dba.session.run_sql("REVOKE SYSTEM_USER ON *.* FROM root ;")',
+            "SET @@SESSION.SQL_LOG_BIN=0;",
+            f"CREATE USER '{self.cluster_admin_user}'@'{self.instance_address}' IDENTIFIED BY '{self.cluster_admin_password}';",
+            f"GRANT ALL ON *.* TO '{self.cluster_admin_user}'@'{self.instance_address}' WITH GRANT OPTION;",
+            f"CREATE USER 'root'@'%' IDENTIFIED BY '{self.root_password}';",
+            "GRANT ALL ON *.* TO 'root'@'%' WITH GRANT OPTION;",
+            "UPDATE mysql.user SET authentication_string=null WHERE User='root' and Host='localhost';",
+            f"ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY '{self.root_password}';",
+            "REVOKE SYSTEM_USER ON *.* FROM root@'%';",
+            "REVOKE SYSTEM_USER ON *.* FROM root@localhost;",
+            "FLUSH PRIVILEGES;",
         )
 
         try:
-            output = self.run_mysqlsh_script("\n".join(_script))
+            output = self._run_mysqlcli_script(" ".join(_script))
             return output.decode("utf-8")
         except subprocess.CalledProcessError as e:
-            logger.exception(f"Failed to configure instance: {self.instance_address}", exc_info=e)
+            logger.exception(f"Failed to configure users for: {self.instance_address}", exc_info=e)
             raise MySQLCreateUserError(e.stdout)
 
     def configure_instance(self):
@@ -101,7 +114,7 @@ class MySQL:
         """Add an instance to the InnoDB cluster."""
         pass
 
-    def run_mysqlsh_script(self, script: str) -> AnyStr:
+    def _run_mysqlsh_script(self, script: str) -> AnyStr:
         """Execute a MySQL shell script.
 
         Raises CalledProcessError if the script gets a non-zero return code.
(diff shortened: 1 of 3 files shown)


src/mysqlsh_helpers.py
```diff
@@ -46,15 +71,36 @@ def mysqlsh_common_dir(self) -> str:
         Returns:
             Path to common dir
         """
-        return "/root/snap/mysql-shell/common"
+        if os.path.exists("/root/snap/mysql-shell/common"):
+            return "/root/snap/mysql-shell/common"
+        else:
```
shayancanonical, line 76: we dont need this `else`:

```
if os.path.exists("/root/snap/mysql-shell/common"):
    return "/root/snap/mysql-shell/common"

return "/tmp"
```

src/mysqlsh_helpers.py
```diff
@@ -46,15 +71,36 @@ def mysqlsh_common_dir(self) -> str:
...
         """Configure the MySQL users for the instance.
 
-        Creates a 'clusteradmin' user with the appropriate privileges and
-        revokes certain privileges from the 'root' user.
+        Creates base `root@%` and `clusteradmin@instance_address` user with the
+        appropriate privileges, and reconfigure `root@localhost` user password.
         """
-        pass
+        _script = (
+            "SET @@SESSION.SQL_LOG_BIN=0;",
```
shayancanonical, line 86: why do we need to disable bin log here? may be add a comment if this is necessary

src/mysqlsh_helpers.py
```diff
@@ -46,15 +71,36 @@ def mysqlsh_common_dir(self) -> str:
...
+            f"CREATE USER '{self.cluster_admin_user}'@'{self.instance_address}' IDENTIFIED BY '{self.cluster_admin_password}';",
+            f"GRANT ALL ON *.* TO '{self.cluster_admin_user}'@'{self.instance_address}' WITH GRANT OPTION;",
+            f"CREATE USER 'root'@'%' IDENTIFIED BY '{self.root_password}';",
+            "GRANT ALL ON *.* TO 'root'@'%' WITH GRANT OPTION;",
+            "UPDATE mysql.user SET authentication_string=null WHERE User='root' and Host='localhost';",
+            f"ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY '{self.root_password}';",
```
shayancanonical, line 92: do we need to reset the password for `root`@`localhost`? also, why use `mysql_native_password`?

src/mysqlsh_helpers.py
```diff
@@ -46,15 +71,36 @@ def mysqlsh_common_dir(self) -> str:
...
+            "REVOKE SYSTEM_USER ON *.* FROM root@localhost;",
+            "FLUSH PRIVILEGES;",
+        )
+
+        try:
+            output = self._run_mysqlcli_script(" ".join(_script))
+            return output.decode("utf-8")
+        except subprocess.CalledProcessError as e:
+            logger.exception(f"Failed to configure users for: {self.instance_address}", exc_info=e)
+            raise MySQLCreateUserError(e.stdout)
```
shayancanonical, line 103: may be rename error to `MySQLConfigureMySQLUsersError`?

src/mysqlsh_helpers.py
```diff
@@ -46,15 +71,36 @@ def mysqlsh_common_dir(self) -> str:
...
+            "UPDATE mysql.user SET authentication_string=null WHERE User='root' and Host='localhost';",
+            f"ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY '{self.root_password}';",
+            "REVOKE SYSTEM_USER ON *.* FROM root@'%';",
+            "REVOKE SYSTEM_USER ON *.* FROM root@localhost;",
+            "FLUSH PRIVILEGES;",
+        )
+
+        try:
+            output = self._run_mysqlcli_script(" ".join(_script))
+            return output.decode("utf-8")
```
shayancanonical, line 100: i'm not sure we should be bubbling the `output` to the caller? the `output` is the return value of `subprocess.check_output`

src/mysqlsh_helpers.py
paulomach: True!

src/mysqlsh_helpers.py
paulomach: This should not be replicated anywhere as it can harm other apps access. Will add a comment.

src/mysqlsh_helpers.py
paulomach: Done as advised by @delgod. AFAIK It's good to have a consistent root password all around, and adding the `mysql_native_password` seems the only way to write the password.

src/mysqlsh_helpers.py
paulomach: Not passionated about that.

src/mysqlsh_helpers.py
paulomach: We can do nothing. I'm not particular with that either. If we convention to suppress it, I'm alright with it.

src/mysqlsh_helpers.py
```diff
@@ -32,6 +55,8 @@ def mysqlsh_bin(self) -> str:
         """
         # Allow for various versions of the mysql-shell snap
         # When we get the alias use /snap/bin/mysqlsh
+        if os.path.exists("/usr/bin/mysqlsh"):
+            return "/usr/bin/mysqlsh"
         if os.path.exists("/snap/bin/mysqlsh"):
             return "/snap/bin/mysqlsh"
         if os.path.exists("/snap/bin/mysql-shell.mysqlsh"):
```
marceloneppel, line 37: Maybe it can be changed to something like the code below (or a lambda function) to avoid repeating the `exist` checks.
```python
for path in ['x', 'y', 'z']:
    if os.path.exists(path)):
        return path
```
But you can ignore it if it makes the code less readable.

marceloneppel (approved):
LGTM!

src/mysqlsh_helpers.py
```diff
@@ -13,15 +13,38 @@
...
+        instance_address: str,
+    ):
+        """Initialize the MySQL class.
+
+        Args:
+            root_password: Password for the 'root' user
+            cluster_admin_user: User name for the cluster admin user
+            cluster_admin_password: Password for the cluster admin user
+            instance_address: Address of the targeted instance
+        """
```
zmraul, line 43: I think this is going to be irrelevant but bear with me for a second :).

The Juju SDK [best practices about docstrings](https://discourse.charmhub.io/t/charm-development-best-practices/5396#heading--docstrings) says that we follow google docstrings style guide, but [google says](https://google.github.io/styleguide/pyguide.html#384-classes) that class atributtes should be declared under the class, not the `__init__`. Then, I took a look around and I saw many other charms doing it the same way as here (uner `init`). So, no conclusions really, but since I spent time to look for this I wanted to leave the note (:

src/mysqlsh_helpers.py
```diff
@@ -46,15 +71,34 @@ def mysqlsh_common_dir(self) -> str:
...
+            return "/root/snap/mysql-shell/common"
+        return "/tmp"
 
     def configure_mysql_users(self):
         """Configure the MySQL users for the instance.
 
-        Creates a 'clusteradmin' user with the appropriate privileges and
-        revokes certain privileges from the 'root' user.
+        Creates base `root@%` and `clusteradmin@instance_address` user with the
+        appropriate privileges, and reconfigure `root@localhost` user password.
```
zmraul, line 82: I would add that it can raise MySQLConfigureMySQLUsersError

zmraul (approved):
Looks good! A small nit and a pedantic one.

src/mysqlsh_helpers.py
```diff
@@ -46,15 +71,34 @@ def mysqlsh_common_dir(self) -> str:
...
         """
-        pass
+        _script = (
+            "SET @@SESSION.SQL_LOG_BIN=0;",
+            f"CREATE USER '{self.cluster_admin_user}'@'{self.instance_address}' IDENTIFIED BY '{self.cluster_admin_password}';",
```
delgod, line 86:
```suggestion
            f"CREATE USER '{self.cluster_admin_user}'@'%' IDENTIFIED BY '{self.cluster_admin_password}';",
```
I think we should have one cluster_admin_user per cluster, not one per instance

delgod (changes requested):

src/mysqlsh_helpers.py
paulomach: @delgod , cluster admin commands should only be run from the operator context, so that's why I've added it in per instance user.
If we do otherwise, one could run cluster admin commands even from client applications. Is that ok?

src/mysqlsh_helpers.py
paulomach: Agreed

src/mysqlsh_helpers.py
paulomach: That's a good point of discussion. I would let it on `__init__` just because the IDE's seem to map it on class instantiation when showing inline doc. But I'm not a strong inclined to either, and I appreciate your dedication.

src/mysqlsh_helpers.py
paulomach: I like this better also.

src/mysqlsh_helpers.py
paulomach: Refactor with [ref](https://docs.oracle.com/cd/E17952_01/mysql-shell-8.0-en/innodb-cluster-user-accounts.html) in mind.

paulomach: Will do on new PR

PR #6, opened 2022-04-07
Refactor MySQL users creations + implement cluster creation and addition of instances to cluster

shayancanonical:
[Jira ticket 1](https://warthogs.atlassian.net/jira/software/c/projects/DPE/boards/390?modal=detail&selectedIssue=DPE-144) and [Jira ticket 2](https://warthogs.atlassian.net/jira/software/c/projects/DPE/boards/390?modal=detail&selectedIssue=DPE-145)

# Issue
1. We are not creating our users for mysql correctly. Additionally, we are only revoking a subset of the required priveleges for the `root` user
2. MySQL is binding to `127.0.0.1` upon installation of `mysql-server-8.0`. However, InnoDB requires cluster instances to be bound to `0.0.0.0`
3. We have not yet implemented the helper method to create a cluster
4. We have not yet implemented the helper method to add an instance to the cluster

# Solution
1. MySQL installation creates `root@localhost`. We additionally create `root@%` and `serverconfig@%` manually. We then create `clusteradmin@%` while configuring the instance for InnoDB (by calling `dba.configure_instance()` within a `mysqlsh` context). Additionally, we will revoke all system user related privileges from `root@localhost` and `root@%`.
See [this link](https://docs.oracle.com/cd/E17952_01/mysql-shell-8.0-en/innodb-cluster-user-accounts.html) for more information on InnoDB users. Also see [this link](https://dev.mysql.com/doc/refman/8.0/en/privileges-provided.html#priv_super) for more information on system user related privileges.

2. Add a mysql configuration template file (`templates/mysqld.cnf`) which defines values for `bind-address` and `mysqlx-bind-address` set to `0.0.0.0`. Copy over these files to the mysql configuration directory (with prefix `z-` so the file has the highest priority) and restart mysql via systemctl during the `on_install` hook.

3. Implement the helper method to create a cluster
4. Implement the helper method to add an instance to a cluster

# Testing
To test, I did the following:
1. Run `charmcraft pack` and deploy a unit using lxd. SSH into the lxd container and ensure that MySQL is running and confirm using the `mysql` client that the bind address values are set to `0.0.0.0`
```
shayan@wellington:~/code/mysql-operator (fix/refactor_user_creation) $ lxc exec juju-30620b-21 bash 
root@juju-30620b-21:~# systemctl status mysql
● mysql.service - MySQL Community Server
     Loaded: loaded (/lib/systemd/system/mysql.service; enabled; vendor preset: enabled)
     Active: active (running) since Thu 2022-04-07 20:11:51 UTC; 35s ago
    Process: 5816 ExecStartPre=/usr/share/mysql/mysql-systemd-start pre (code=exited, status=0/SUCCESS)
   Main PID: 5824 (mysqld)
     Status: "Server is operational"
      Tasks: 38 (limit: 18730)
     Memory: 357.9M
     CGroup: /system.slice/mysql.service
             └─5824 /usr/sbin/mysqld

Apr 07 20:11:50 juju-30620b-21 systemd[1]: Starting MySQL Community Server...
Apr 07 20:11:51 juju-30620b-21 systemd[1]: Started MySQL Community Server.
root@juju-30620b-21:~# mysql

mysql> show variables like '%addr%';
+---------------------+---------+
| Variable_name       | Value   |
+---------------------+---------+
| admin_address       |         |
| bind_address        | 0.0.0.0 |
| mysqlx_bind_address | 0.0.0.0 |
+---------------------+---------+
3 rows in set (0.00 sec)
```

2. I launched 2 units with juju using lxd. Then SSH into both units, navigate to the juju source code directory, instantiate an instance of `MySQL` helper class and run the `configure_mysql_users` and `configure_instance` on both lxd containers. Then, on one container run `create_cluster()` and `add_instance_to_cluster()` methods.

# Release Notes
- Implement the helper functions to configure mysql, create the appropriate mysql users, configure the instance, create a cluster and add instances to the cluster.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -61,13 +61,13 @@ class MySQLOperatorCharm(CharmBase):
             # TODO: replace stubbed arguments once mechanisms to generate them exist
             # Mechanisms = generating user/pass and storing+retrieving them from peer databag.
             self._mysql_helpers = MySQL(
-                "clusteradminpassword",
-                "clusteradmin",
-                "test_cluster",
                 "127.0.0.1",
+                "test_cluster",
                 "password",
-                "serverconfigpassword",
                 "serverconfig",
+                "serverconfigpassword",
+                "clusteradmin",
+                "clusteradminpassword",
             )
 
         return self._mysql_helpers
--- a/src/charm.py
+++ b/src/charm.py
@@ -5,25 +5,14 @@
 """Charmed Machine Operator for MySQL."""
 
 import logging
-import subprocess
 
-from charms.operator_libs_linux.v0 import apt
-from charms.operator_libs_linux.v1 import snap
 from ops.charm import CharmBase
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus
 
-from mysqlsh_helpers import (
-    MySQL,
-    MySQLInitializationError,
-    MySQLUpdateConfigurationError,
-)
+from mysqlsh_helpers import MySQL
 
 logger = logging.getLogger(__name__)
-
-# TODO: determine if version locking is needed for both mysql-shell and mysql-server
-MYSQL_SHELL_SNAP_NAME = "mysql-shell"
-MYSQL_APT_PACKAGE_NAME = "mysql-server-8.0"
 
 
 class MySQLOperatorCharm(CharmBase):
@@ -46,47 +35,10 @@ class MySQLOperatorCharm(CharmBase):
         """Initial setup operations like installing dependencies, and creating users and groups."""
         self.unit.status = MaintenanceStatus("Installing MySQL")
 
-        # Install 'mysql-server-8.0' apt package
-        # Note: installing mysql-server will create the 'mysql' user and 'mysql' group,
-        # and run mysql with the 'mysql' user
         try:
-            logger.debug("Updating apt cache")
-            apt.update()
-        except subprocess.CalledProcessError as e:
-            logger.exception("Failed to update apt cache", exc_info=e)
-            self.unit.status = BlockedStatus("Failed to update apt")
-            return
-
-        try:
-            logger.debug(f"Installing '{MYSQL_APT_PACKAGE_NAME}' apt package")
-            apt.add_package(MYSQL_APT_PACKAGE_NAME)
-        except (apt.PackageNotFoundError, apt.PackageError) as e:
-            logger.exception(
-                f"Failed to install '{MYSQL_APT_PACKAGE_NAME}' apt package",
-                exc_info=e,
-            )
-            self.unit.status = BlockedStatus(f"Failed to install '{MYSQL_APT_PACKAGE_NAME}'")
-            return
-
-        # Install 'mysql-shell' snap
-        try:
-            cache = snap.SnapCache()
-            mysql_shell = cache[MYSQL_SHELL_SNAP_NAME]
-
-            if not mysql_shell.present:
-                logger.debug(f"Installing '{MYSQL_SHELL_SNAP_NAME}' snap")
-                mysql_shell.ensure(snap.SnapState.Latest, channel="stable")
-        except (snap.SnapNotFoundError, snap.SnapError) as e:
-            logger.exception(f"Failed to install the '{MYSQL_SHELL_SNAP_NAME}' snap", exc_info=e)
-            self.unit.status = BlockedStatus(f"Failed to install '{MYSQL_SHELL_SNAP_NAME}'")
-            return
-
-        # Update the mysql configuration - from templates/mysqld.cnf
-        try:
-            mysql_helpers = self._get_mysql_helpers()
-            mysql_helpers.update_mysql_configuration()
-        except (MySQLInitializationError, MySQLUpdateConfigurationError):
-            self.unit.status = BlockedStatus("Failed to update the mysql configuration")
+            MySQL.install_and_configure_mysql_dependencies()
+        except Exception:
+            self.unit.status = BlockedStatus("Failed to install and configure MySQL")
             return
 
         # TODO: Set status to WaitingStatus once _on_start is implemented
@@ -102,7 +54,8 @@ class MySQLOperatorCharm(CharmBase):
     #  Helpers
     # =======================
 
-    def _get_mysql_helpers(self):
+    @property
+    def _mysql(self):
         """Returns an instance of the MySQL object from mysqlsh_helpers."""
         if not self._mysql_helpers:
             # TODO: replace stubbed arguments once mechanisms to generate them exist
--- a/src/charm.py
+++ b/src/charm.py
@@ -21,9 +21,6 @@ class MySQLOperatorCharm(CharmBase):
     def __init__(self, *args):
         super().__init__(*args)
 
-        # Please do not reference this variable directly. Instead use _get_mysql_helpers().
-        self._mysql_helpers = None
-
         self.framework.observe(self.on.install, self._on_install)
         self.framework.observe(self.on.start, self._on_start)
 
@@ -54,23 +51,19 @@ class MySQLOperatorCharm(CharmBase):
     #  Helpers
     # =======================
 
-    @property
-    def _mysql(self):
+    def _get_mysql_helpers(self):
         """Returns an instance of the MySQL object from mysqlsh_helpers."""
-        if not self._mysql_helpers:
-            # TODO: replace stubbed arguments once mechanisms to generate them exist
-            # Mechanisms = generating user/pass and storing+retrieving them from peer databag.
-            self._mysql_helpers = MySQL(
-                "127.0.0.1",
-                "test_cluster",
-                "password",
-                "serverconfig",
-                "serverconfigpassword",
-                "clusteradmin",
-                "clusteradminpassword",
-            )
-
-        return self._mysql_helpers
+        # TODO: replace stubbed arguments once mechanisms to generate them exist
+        # Mechanisms = generating user/pass and storing+retrieving them from peer databag.
+        return MySQL(
+            "127.0.0.1",
+            "test_cluster",
+            "password",
+            "serverconfig",
+            "serverconfigpassword",
+            "clusteradmin",
+            "clusteradminpassword",
+        )
 
 
 if __name__ == "__main__":
--- a/src/charm.py
+++ b/src/charm.py
@@ -32,8 +32,8 @@ class MySQLOperatorCharm(CharmBase):
     def __init__(self, *args):
         super().__init__(*args)
 
-        # initialized in _on_install()
-        self.mysql_helpers = None
+        # Please do not reference this variable directly. Instead use _get_mysql_helpers().
+        self._mysql_helpers = None
 
         self.framework.observe(self.on.install, self._on_install)
         self.framework.observe(self.on.start, self._on_start)
@@ -83,17 +83,8 @@ class MySQLOperatorCharm(CharmBase):
 
         # Update the mysql configuration - from templates/mysqld.cnf
         try:
-            # TODO: replace stubbed arguments once mechanisms to generate them exist
-            self.mysql_helpers = MySQL(
-                "clusteradminpassword",
-                "clusteradmin",
-                "test_cluster",
-                "127.0.0.1",
-                "password",
-                "serverconfigpassword",
-                "serverconfig",
-            )
-            self.mysql_helpers.update_mysql_configuration()
+            mysql_helpers = self._get_mysql_helpers()
+            mysql_helpers.update_mysql_configuration()
         except (MySQLInitializationError, MySQLUpdateConfigurationError):
             self.unit.status = BlockedStatus("Failed to update the mysql configuration")
             return
@@ -106,6 +97,27 @@ class MySQLOperatorCharm(CharmBase):
     def _on_start(self, _) -> None:
         """Ensure that required software is running."""
         pass
+
+    # =======================
+    #  Helpers
+    # =======================
+
+    def _get_mysql_helpers(self):
+        """Returns an instance of the MySQL object from mysqlsh_helpers."""
+        if not self._mysql_helpers:
+            # TODO: replace stubbed arguments once mechanisms to generate them exist
+            # Mechanisms = generating user/pass and storing+retrieving them from peer databag.
+            self._mysql_helpers = MySQL(
+                "clusteradminpassword",
+                "clusteradmin",
+                "test_cluster",
+                "127.0.0.1",
+                "password",
+                "serverconfigpassword",
+                "serverconfig",
+            )
+
+        return self._mysql_helpers
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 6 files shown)


src/charm.py
```diff
@@ -74,15 +76,19 @@ def _on_install(self, _) -> None:
...
-            return
-        except snap.SnapError as e:
+        except (snap.SnapNotFoundError, snap.SnapError) as e:
             logger.exception(f"Failed to install the '{MYSQL_SHELL_SNAP_NAME}' snap", exc_info=e)
             self.unit.status = BlockedStatus(f"Failed to install '{MYSQL_SHELL_SNAP_NAME}'")
             return
 
+        # Update the mysql configuration - from templates/mysqld.cnf
+        try:
+            mysql_helpers = self._get_mysql_helpers()
```
shayancanonical, line 86: should call `mysqlsh --help` after the installation of `mysql-shell` (to create mysql shell common directories) and avoid the `MySQL._ensure_mysqlsh_common_dir()` method

src/mysqlsh_helpers.py
```diff
@@ -35,84 +61,143 @@ class MySQL:
...
-        return "/tmp"
+        return "/root/snap/mysql-shell/common"
+
+    def update_mysql_configuration(self):
+        """Add a configuration file for mysqld and restart the mysql service."""
+        try:
+            # target file starts with 'z-' so it has priority over the default config file
+            shutil.copyfile("templates/mysqld.cnf", "/etc/mysql/mysql.conf.d/z-custom-mysqld.cnf")
+
+            restart_mysql_command = ["systemctl", "restart", "mysql"]
```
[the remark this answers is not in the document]
paulomach: +1 to that. Much cleaner with the lib.

src/mysqlsh_helpers.py
```diff
@@ -35,84 +61,143 @@ class MySQL:
...
-            "REVOKE SYSTEM_USER ON *.* FROM root@'%';",
-            "REVOKE SYSTEM_USER ON *.* FROM root@localhost;",
+            f"REVOKE {', '.join(privileges_to_revoke)} ON *.* FROM root@'%';",
+            f"REVOKE {', '.join(privileges_to_revoke)} ON *.* FROM root@localhost;",
             "FLUSH PRIVILEGES;",
         )
 
         try:
-            self._run_mysqlcli_script(" ".join(_script))
+            logger.debug("Configuring MySQL users=")
```
paulomach, line 181: Is this `=` on purpose on the log message?

src/mysqlsh_helpers.py
```diff
@@ -123,15 +208,56 @@ def configure_instance(self) -> None:
...
+
+        Raises MySQLADDInstanceToClusterError
+            if there was an issue adding the instance to the cluster.
+
+        Args:
+            instance_address: address of the instance to add to the cluster
+        """
+        options = {
+            "password": self.cluster_admin_password,
+            "recoveryMethod": "clone",
```
paulomach, line 243: Should it always `clone`? In the case of a re-adding an instance, can this be potentially very taxing? What's the thoughts against leaving it in `auto` ?

src/mysqlsh_helpers.py
```diff
@@ -35,84 +61,143 @@ class MySQL:
...
+        commands = (
             "SET @@SESSION.SQL_LOG_BIN=0;",
-            f"CREATE USER '{self.cluster_admin_user}'@'{self.instance_address}' IDENTIFIED BY '{self.cluster_admin_password}';",
-            f"GRANT ALL ON *.* TO '{self.cluster_admin_user}'@'{self.instance_address}' WITH GRANT OPTION;",
             f"CREATE USER 'root'@'%' IDENTIFIED BY '{self.root_password}';",
             "GRANT ALL ON *.* TO 'root'@'%' WITH GRANT OPTION;",
+            f"CREATE USER '{self.server_config_user}'@'%' IDENTIFIED BY '{self.server_config_password}';",
+            f"GRANT ALL ON *.* TO '{self.server_config_user}'@'%' WITH GRANT OPTION;",
             "UPDATE mysql.user SET authentication_string=null WHERE User='root' and Host='localhost';",
             f"ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY '{self.root_password}';",
```
delgod, line 174: please don't use `mysql_native_password`! it is unsafe, you can read my article https://www.percona.com/blog/2020/06/12/brute-force-mysql-password-from-a-hash/

delgod (changes requested):

src/mysqlsh_helpers.py
```diff
@@ -35,84 +61,143 @@ class MySQL:
...
-        if os.path.exists("/root/snap/mysql-shell/common"):
-            return "/root/snap/mysql-shell/common"
-        return "/tmp"
+        return "/root/snap/mysql-shell/common"
+
+    def update_mysql_configuration(self):
+        """Add a configuration file for mysqld and restart the mysql service."""
+        try:
+            # target file starts with 'z-' so it has priority over the default config file
+            shutil.copyfile("templates/mysqld.cnf", "/etc/mysql/mysql.conf.d/z-custom-mysqld.cnf")
```
delgod, line 134: why not put the file before starting installing the mysqld?
some options change the binary format and can be set only before the creation of MySQL dir

src/charm.py
paulomach: I say to let this kind of domain into the helper, so charm code deals with charming tasks.

src/mysqlsh_helpers.py
```diff
@@ -140,11 +266,22 @@ def _wait_until_mysql_connection(self) -> None:
         Retry every 5 seconds for 30 seconds if there is an issue obtaining a connection.
         """
         commands = (
-            f"my_shell = shell.connect('{self.cluster_admin_user}:{self.cluster_admin_password}@{self.instance_address}')",
+            f"my_shell = shell.connect('root:{self.root_password}@{self.instance_address}')",
```
delgod, line 269: it is simpler to check if socket file was created - such check don't need auth

src/charm.py
```diff
@@ -92,6 +98,27 @@ def _on_start(self, _) -> None:
         """Ensure that required software is running."""
         pass
 
+    # =======================
+    #  Helpers
+    # =======================
+
+    def _get_mysql_helpers(self):
```
delgod, line 105: why not have this function as a property?
in such case, it will be possible to use it as `self._mysql.update_mysql_configuration()`

src/charm.py
```diff
@@ -92,6 +98,27 @@ def _on_start(self, _) -> None:
...
+        """Returns an instance of the MySQL object from mysqlsh_helpers."""
+        if not self._mysql_helpers:
+            # TODO: replace stubbed arguments once mechanisms to generate them exist
```
paulomach, line 108: Add to this TODO that the mechanism must check for credentials existence in the peer relationship data bag.
@shayancanonical , thoughts on this being the nexts tasks?

delgod (changes requested):

paulomach (commented):
In general, all seems good, but I want to sync about some of the decisions before going any further.

src/mysqlsh_helpers.py
paulomach: Can't the socket survive a daemon crash?

src/charm.py
paulomach: I don't disagree, and if we extrapolate this path, we would create another helper to deal with all installation duties.

src/mysqlsh_helpers.py
paulomach: That's on me.

src/mysqlsh_helpers.py
shayancanonical: confirmed that the socket file does not exist if the service is stopped. will be changing to check if socket exists in future commit

src/mysqlsh_helpers.py
shayancanonical: I tried to set to `auto`, but got the following error trace:

```
The target instance \'juju-840aa8-2:3306\' has not been pre-provisioned (GTID set is empty). The Shell is unable to decide whether incremental state recovery can correctly provision it.
The safest and most convenient way to provision a new instance is through automatic clone provisioning, which will completely overwrite the state of \'juju-840aa8-2:3306\' with a physical snapshot from an existing cluster member. To use this method by default, set the \'recoveryMethod\' option to \'clone\'.

The incremental state recovery may be safely used if you are sure all updates ever executed in the cluster were done with GTIDs enabled, there are no purged transactions and the new instance contains the same GTID set as the cluster or a subset of it. To use this method by default, set the \'recoveryMethod\' option to \'incremental\'.

Traceback (most recent call last):
  File "<string>", line 3, in <module>
RuntimeError: Cluster.add_instance: \'recoveryMethod\' option must be set to \'clone\' or \'incremental\'
```

src/mysqlsh_helpers.py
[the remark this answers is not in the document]
shayancanonical: didnt need the systemd lib, since we now add the custom mysqld config before downloading mysql-server (thus we no longer need to restart mysql)

src/charm.py
shayancanonical: moved all install actions into a static method in MySQL helper class

src/mysqlsh_helpers.py
shayancanonical: removed in 37db8b57689305a169c643d92613f15416ef8710

src/mysqlsh_helpers.py
shayancanonical: @delgod @paulomach i removed usage of `mysql_native_password`. instead (in 37db8b57689305a169c643d92613f15416ef8710), mysql is invoked twice (once to create the `root@%` user with a password, and the second to configure all users using the password for `root@%` when calling `mysql`). please confirm if [this](https://github.com/canonical/mysql-operator/pull/6/files#diff-6e566a7b93efc3f3e7ac0e6d8989256725f876f06af78c45d0ae3815d4bcf82eR189-R213) is a good approach

src/charm.py
shayancanonical: yes! we should definitely tackle the charm related logic next!

src/mysqlsh_helpers.py
shayancanonical: we now copy the config file before installing mysql-server in [37db8b5](https://github.com/canonical/mysql-operator/pull/6/commits/37db8b57689305a169c643d92613f15416ef8710)

src/mysqlsh_helpers.py
shayancanonical: changed to check if socket file exists in [37db8b5](https://github.com/canonical/mysql-operator/pull/6/commits/37db8b57689305a169c643d92613f15416ef8710)

src/charm.py
shayancanonical: changed to a property in [37db8b5](https://github.com/canonical/mysql-operator/pull/6/commits/37db8b57689305a169c643d92613f15416ef8710) whose usage will be further formalized in the future

src/mysqlsh_helpers.py
paulomach: That's interesting. @delgod illuminate us with your thoughts

paulomach (approved):
LGTM, I still want to know about the always clone on add instance.

src/charm.py
```diff
@@ -26,6 +21,9 @@ class MySQLOperatorCharm(CharmBase):
     def __init__(self, *args):
         super().__init__(*args)
 
+        # Please do not reference this variable directly. Instead use _get_mysql_helpers().
+        self._mysql_helpers = None
+
```
marcoppenheimer, line 26: Perhaps I'm slow, what's the purpose of this line if we have the property below?

If we removed the `if` from the `_mysql()` property, always returning a new instance of `MySQL` what would be the impact?

src/mysqlsh_helpers.py
```diff
@@ -35,84 +66,172 @@ class MySQL:
...
+        self.server_config_user = server_config_user
+        self.server_config_password = server_config_password
         self.cluster_admin_user = cluster_admin_user
         self.cluster_admin_password = cluster_admin_password
-        self.instance_address = instance_address
 
-    @property
-    def mysqlsh_bin(self) -> str:
+    @staticmethod
+    def _get_mysqlsh_bin() -> str:
```
marcoppenheimer, line 97: Again I could be wrong, but the pattern of underscore-prefixing object methods is usually to indicate that this method shouldn't be used from outside the object. 

Whereas a `staticmethod` tells me that it's also able to be used from outside the object.

src/mysqlsh_helpers.py
shayancanonical: Nope, it was just a typo, but thanks for the link - I did not know you could do that

src/charm.py
shayancanonical: The `_mysql()` property will in a future PR retrieve values from the peer relation databag to instantiate a new instance of `MySQL` (inputs to the constructor are currently stubbed). The idea behind caching in the object is to avoid retrieving from the databag over again. However, now that you point this out, may be it's a good idea not to cache and to instead re-instantiate (I dont imagine that retrieving from the peer databag is too expensive)

src/mysqlsh_helpers.py
shayancanonical: You are right! Somehow, my reasoning was that it's a staticmethod only used within the context of the object (and so it's private). But there is no such thing!

Updated to public in [7dbb768](https://github.com/canonical/mysql-operator/pull/6/commits/7dbb7684b4aab2198d0bdae74307b83f093fbb8d)

src/mysqlsh_helpers.py
delgod: it is ok

src/mysqlsh_helpers.py
delgod: I think we should try `auto` first. and switch to `clone` only if `auto` doesn't work

delgod (approved):

src/mysqlsh_helpers.py
shayancanonical: addressed in [aec0bd5](https://github.com/canonical/mysql-operator/pull/6/commits/aec0bd5f1ba0151b639a94eb6959bbc4ed6c7349). we try `auto` first, and if there is an exception, we try `clone`

src/mysqlsh_helpers.py
```diff
@@ -278,24 +278,38 @@ def add_instance_to_cluster(self, instance_address) -> None:
         """
         options = {
             "password": self.cluster_admin_password,
-            "recoveryMethod": "clone",
+            "recoveryMethod": "auto",
```
marceloneppel, line 281: I guess this can be removed as it's being overriden on L291.

marceloneppel (approved):
LGTM! Thanks for the details on the testing part.

src/mysqlsh_helpers.py
shayancanonical: good catch. removed in 27779e4d85e1bc2ab3ce6b2d5b27e9078fd09b99

PR #7, opened 2022-04-12
Generate and store MySQL config values in the peer relation databag

shayancanonical:
[Jira ticket](https://warthogs.atlassian.net/jira/software/c/projects/DPE/boards/390?modal=detail&selectedIssue=DPE-206)

# Issue
We have stubbed MySQL config values in the charm.

# Solution
Generate the MySQL config values upon first execution and store values in the peer relation databag. Upon future executions, retrieve values from the databag.

The following values are stored in the databag:
- cluster_name: derived either from a config value or a randomly generated hash
- root_password: randomly generated
- server_config_password: randomly generated
- cluster_admin_password: randomly generated

# Release Notes
- Generate and store+retrieve MySQL config values in/from the peer relation databag.


config.yaml
```diff
@@ -0,0 +1,7 @@
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+options:
+  cluster_name:
+    description: "Name of the MySQL InnoDB cluster"
+    type: "string"
```
shayancanonical, line 7: new line

src/charm.py
```diff
@@ -4,24 +4,50 @@
...
+from mysqlsh_helpers import (
+    MySQL,
+    MySQLConfigureInstanceError,
+    MySQLConfigureMySQLUsersError,
+)
 
 logger = logging.getLogger(__name__)
 
+PASSWORD_LENGTH = 24
+PEER = "mysql-replicas"
```
shayancanonical, line 25: read from meta file?

paulomach: I say, keep on code, which is what everyone else is doing.

src/charm.py
```diff
@@ -58,20 +93,60 @@ def _on_start(self, _) -> None:
...
                 "clusteradmin",
-                "clusteradminpassword",
+                mysql_configs["cluster_admin_password"],
             )
 
         return self._mysql_helpers
 
+    def _get_or_create_mysql_configs(self):
+        peer_relation = self.model.get_relation(PEER)
+        if peer_relation is None:
```
paulomach, line 112: nit: test with `not peer_relation`

src/charm.py
```diff
@@ -4,24 +4,50 @@
...
+    MySQLConfigureMySQLUsersError,
+)
 
 logger = logging.getLogger(__name__)
 
+PASSWORD_LENGTH = 24
+PEER = "mysql-replicas"
+
+
+def generate_random_password(length: int) -> str:
```
paulomach, line 28: Wonder if we should put these two on the helpers

src/charm.py
```diff
@@ -58,20 +93,60 @@ def _on_start(self, _) -> None:
...
+            "unit_ip": self.model.get_binding(PEER).network.bind_address,
+            "cluster_name": peer_relation.data[self.app].get("cluster_name"),
+            "root_password": peer_relation.data[self.app].get("root_password"),
+            "server_config_password": peer_relation.data[self.app].get("server_config_password"),
+            "cluster_admin_password": peer_relation.data[self.app].get("cluster_admin_password"),
+        }
+
+        is_unit_leader = self.unit.is_leader()
+        has_missing_peer_data = any([value is None for value in mysql_configs.values()])
+        if has_missing_peer_data and not is_unit_leader:
```
paulomach, line 125: When raising this exception, it will not get threated in the hook code. Should'n this be mapped to new Exceptions and threated accordingly? (even though the hooks also test for leadership.)

paulomach (changes requested):
Overall liked the approach on access to helper class, with some minor cr's

src/charm.py
shayancanonical: I was thinking that `mysqlsh_helpers` will only contain helpers for mysql. these are functions at the moment, because i expect them to be replaced with `data-platform-libs` in the future

config.yaml
shayancanonical: added in 516bd9596e3be0e50d16a526485b3a330a60f2ff

src/charm.py
shayancanonical: check no longer needed - it was removed during the refactor in 516bd9596e3be0e50d16a526485b3a330a60f2ff

src/charm.py
shayancanonical: i was originally thinking that these exceptions would push the charm into an error state as these conditions should never occur. however, with the refactor (516bd9596e3be0e50d16a526485b3a330a60f2ff), we can make the assumption that the data will be available during the `start` event

src/charm.py
```diff
@@ -58,20 +93,60 @@ def _on_start(self, _) -> None:
...
+            raise Exception("Trying to store data in peer relation on non-leader unit")
+
+        if not mysql_configs["cluster_name"]:
```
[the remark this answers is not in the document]
shayancanonical: added a variation of this suggestion in 516bd9596e3be0e50d16a526485b3a330a60f2ff

marceloneppel (approved):
Looks good!

paulomach (approved):
LGTM !

shayancanonical:
Thanks for the reviews!

PR #8, opened 2022-04-18
Modify on_start to create a cluster on the leader unit + minor bugfixes

shayancanonical:
# Issue
- We are only configuring the instance for the leader unit
- We are not attaching labels to instances (and so units in the cluster appear by their `hostname`s). This would cause problems trying to query the status of an instance using mysqlsh since we do not have access to the hostname in the ops framework
- Config keys cannot have `_`s
- The cluster name cannot have characters other than letters, digits and `_`s. Also, the naming is not consistent with the k8s charm

# Solution
- Configuring all instances, but create cluster only on the leader unit
- Attach labels (unit names) to each instance in the cluster (would appear as `mysql-0` instead of `juju-<hash>-<unit_number>`)
- Use `-`s instead of `_` in config keys and peer relation data keys
- Randomly generate a hash for the cluster name if one is not specified in the config (will look like `cluster_<md5_hash>`

# Release Notes
- Minor updates to get a single node cluster up and running in the charm

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -111,6 +111,8 @@ class MySQLOperatorCharm(CharmBase):
 
         # Create the cluster on the juju leader unit
         if not self.unit.is_leader():
+            # TODO: change to WaitingStatus, and move to ActiveStatus in the update-status
+            # event handler after the instance has been added to the cluster
             self.unit.status = ActiveStatus()
             return
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -17,6 +17,7 @@ from mysqlsh_helpers import (
     MySQL,
     MySQLConfigureInstanceError,
     MySQLConfigureMySQLUsersError,
+    MySQLCreateClusterError,
 )
 
 logger = logging.getLogger(__name__)
@@ -35,9 +36,9 @@ def generate_random_password(length: int) -> str:
     return "".join([secrets.choice(choices) for i in range(length)])
 
 
-def generate_random_hash() -> str:
-    """Generate a hash based on a random string."""
-    random_characters = generate_random_password(10)
+def generate_random_hash(length=20) -> str:
+    """Generate a random hash."""
+    random_characters = generate_random_password(20)
     return hashlib.md5(random_characters.encode("utf-8")).hexdigest()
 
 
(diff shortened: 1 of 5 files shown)


WRFitch (approved):
LGTM

marceloneppel (approved):
Looks great!

src/charm.py
```diff
@@ -90,23 +91,36 @@ def _on_config_changed(self, _) -> None:
...
         except MySQLConfigureMySQLUsersError:
             self.unit.status = BlockedStatus("Failed to initialize MySQL users")
             return
         except MySQLConfigureInstanceError:
             self.unit.status = BlockedStatus("Failed to configure instance for InnoDB")
             return
 
+        # Create the cluster on the juju leader unit
+        if not self.unit.is_leader():
+            self.unit.status = ActiveStatus()
```
paulomach, line 114: Common lifecycle question here:
You're setting non leader unit as Active, but the unit at this point is not a cluster member.
Wouldn't be more correct to set it to WaitingStatus (e.g. "waiting for peer relation join")?

And then, there's another question: since the leader unit is the one which join the incoming non-leader unit to the cluster, how to update it status from waiting->active, after a successful add_instance that was done by the leader?

src/mysqlsh_helpers.py
```diff
@@ -336,7 +339,7 @@ def _run_mysqlsh_script(self, script: str) -> None:
             # Specify python as this is not the default in the deb version
             # of the mysql-shell snap
             command = [MySQL.get_mysqlsh_bin(), "--no-wizard", "--python", "-f", _file.name]
-            subprocess.check_output(command, stderr=subprocess.PIPE)
+            return subprocess.check_output(command, stderr=subprocess.PIPE).decode("utf-8")
```
paulomach, line 342: Why add the return? I couldn't see where is being used.

paulomach (approved):
LGTM. just need some clarifications to understand decisions made.

src/charm.py
shayancanonical: Discussed offline with Paulo: non-leader units should be moved into `WaitingStatus`. And then, in the `update-status` event handler, we'd move the unit to `ActiveStatus` after confirming that it is part of the InnoDB cluster.

We will implement this in a future PR - added a TODO item in fc409d658a9d226f29dd0b36a2723615e638dd98

src/mysqlsh_helpers.py
shayancanonical: Ah, this should have been introduced in a future PR. Removed for the time being in fc409d658a9d226f29dd0b36a2723615e638dd98

PR #9, opened 2022-04-25
Implement scaling up and down of the InnoDB cluster

shayancanonical:
[Jira ticket](https://warthogs.atlassian.net/jira/software/c/projects/DPE/boards/390?modal=detail&selectedIssue=DPE-215)

# Issue
- We can only provision a cluster of one instance (we have not yet implemented the inclusion of additional instances to the InnoDB cluster)
- We are not attaching any storage volumes in the charm's metadata
- We are unable to scale down the cluster correctly

# Solution
- Add instances to the cluster on the leader unit's `on_<peer>_relation_joined` event
- Attach a storage volume for each unit at path `/var/lib/mysql/` (where MySQL's data is stored)
- Scale down in `storage_detaching` event (utilizing locks provided by the underlying database to ensure that only one instance is scaled down at a time - a requirement to avoid leaving the cluster in a bad state).

# Context
To ensure that we only remove one instance at a time from the cluster (to avoid split-brain or lack of majority issues), we create a `mysql.juju_units_operations` table to which we [update+select](https://github.com/canonical/mysql-operator/pull/9/files#diff-6e566a7b93efc3f3e7ac0e6d8989256725f876f06af78c45d0ae3815d4bcf82eR431-R436) to acquire a lock, and [update](https://github.com/canonical/mysql-operator/pull/9/files#diff-6e566a7b93efc3f3e7ac0e6d8989256725f876f06af78c45d0ae3815d4bcf82eR487-R490) again to release the lock.

# Testing
More comprehensive integration tests will be added in a followup PR

# Release Notes
- Enable the scale up and scale down of the MySQL InnoDB cluster

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -16,5 +16,5 @@ maintainers:
   - Shayan Patel <shayan.patel@canonical.com>
 
 peers:
-  mysql-replicas:
-    interface: mysql-replicas
+  database-peers:
+    interface: mysql-peers
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -9,16 +9,16 @@ summary: |
   MySQL is  a widely used, open-source relational database management system
   (RDBMS). MySQL InnoDB cluster provides a complete high availability solution
   for MySQL via Group Replication.
+source: https://github.com/canonical/mysql-operator
+issues: https://github.com/canonical/mysql-operator/issues
 
   This charm supports MySQL 8 on machines.
 maintainers:
   - Paulo Machado <paulo.machado@canonical.com>
   - Shayan Patel <shayan.patel@canonical.com>
-
 peers:
   database-peers:
     interface: mysql-peers
-
 storage:
   database:
     type: filesystem
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -18,3 +18,9 @@ maintainers:
 peers:
   database-peers:
     interface: mysql-peers
+
+storage:
+  database:
+    type: filesystem
+    description: Persistent storage for MySQL data
+    location: /var/lib/mysql
--- a/src/mysqlsh_helpers.py
+++ b/src/mysqlsh_helpers.py
@@ -34,6 +34,8 @@ MYSQL_SHELL_COMMON_DIRECTORY = "/root/snap/mysql-shell/common"
 MYSQLD_SOCK_FILE = "/var/run/mysqld/mysqld.sock"
 MYSQLD_CONFIG_DIRECTORY = "/etc/mysql/mysql.conf.d"
 
+UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
+
 
 class MySQLConfigureMySQLUsersError(Exception):
     """Exception raised when creating a user fails."""
@@ -71,13 +73,20 @@ class MySQLConfirmAllInstancesOnlineError(Exception):
     pass
 
 
-class MySQLRemoveInstanceError(Exception):
+class MySQLRemoveInstanceRetryError(Exception):
     """Exception raised when there is an issue removing an instance.
 
     Utilized by tenacity to retry the method.
     """
 
     pass
+
+
+class MySQLRemoveInstanceError(Exception):
+    """Exception raised when there is an issue removing an instance.
+
+    Exempt from the retry mechanism provided by tenacity.
+    """
 
 
 class MySQLInitializeJujuOperationsTableError(Exception):
@@ -299,7 +308,7 @@ class MySQL:
         """Initialize the mysql.juju_units_operations table using the serverconfig user."""
         initalize_table_commands = (
             "CREATE TABLE mysql.juju_units_operations (task varchar(20), executor varchar(20), status varchar(20), primary key(task));",
-            "INSERT INTO mysql.juju_units_operations values ('unit-teardown', '', 'not-started');",
+            f"INSERT INTO mysql.juju_units_operations values ('{UNIT_TEARDOWN_LOCKNAME}', '', 'not-started');",
         )
 
         try:
@@ -400,7 +409,7 @@ class MySQL:
             return False
 
     @retry(
-        retry=retry_if_exception_type(MySQLRemoveInstanceError),
+        retry=retry_if_exception_type(MySQLRemoveInstanceRetryError),
         stop=stop_after_attempt(15),
         reraise=True,
         wait=wait_random(min=4, max=30),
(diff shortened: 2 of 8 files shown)


src/mysqlsh_helpers.py
```diff
@@ -311,6 +369,197 @@ def add_instance_to_cluster(self, instance_address, instance_unit_label) -> None
...
+            )
+            self._run_mysqlsh_script("\n".join(remove_instance_commands))
+
+            # There is no need to release the lock if cluster was dissolved
+            if not remaining_cluster_member_addresses:
+                return
+
+            # Retrieve the cluster primary's address again (in case the old primary is scaled down)
+            # Release the lock by making a request to this primary member's address
+            primary_address = self._get_cluster_primary_address(
```
paulomach, line 480: should it have some retry logic in case this calls returns None?

paulomach (approved):
@shayancanonical , I've added just one concerns on the removal process. 
Looks really nice, so much new stuff to use on k8s ;)

metadata.yaml
```diff
@@ -16,5 +16,11 @@ maintainers:
...
-  mysql-replicas:
-    interface: mysql-replicas
+  database-peers:
+    interface: mysql-peers
+
+storage:
+  database:
+    type: filesystem
+    description: Persistent storage for MySQL data
+    location: /var/lib/mysql
```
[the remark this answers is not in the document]
shayancanonical: addressed in 0a55b4d219288fca33ec9f0673995488e198f445

src/mysqlsh_helpers.py
```diff
@@ -283,9 +305,12 @@ def add_instance_to_cluster(self, instance_address, instance_unit_label) -> None
 
         connect_commands = (
             f"shell.connect('{self.cluster_admin_user}:{self.cluster_admin_password}@{self.instance_address}')",
+            "session.run_sql(\"SELECT get_lock('add_instance', -1);\")",
```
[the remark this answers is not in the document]
shayancanonical: great catch! we do not need a lock here - this is a relic of testing that should have never have been commit. removed in 0a55b4d219288fca33ec9f0673995488e198f445

src/mysqlsh_helpers.py
shayancanonical: sync'ed offline - we should not be retrying any operations after the successful removal of the instance from the cluster. moved the block following the instance removal out of the `try/except`, and added a condition to raise an exception if the retrieval of the primary address failed in 0a55b4d219288fca33ec9f0673995488e198f445

src/mysqlsh_helpers.py
```diff
@@ -311,6 +367,200 @@ def add_instance_to_cluster(self, instance_address, instance_unit_label) -> None
...
+
+            output = self._run_mysqlsh_script("\n".join(acquire_lock_commands))
+            matches = re.search(r"<ACQUIRED_LOCK>(\d)</ACQUIRED_LOCK>", output)
+            if not matches:
+                raise MySQLRemoveInstanceError("Unable to acquire lock to remove unit")
+
+            # If unable to acquire the lock, raise an error to retry
+            acquired_lock = int(matches.group(1))
+            if not acquired_lock:
+                raise MySQLRemoveInstanceError("Did not acquire lock to remove unit")
```
WRFitch, line 444: This method's pretty big - would it makes sense to refactor out the lock logic to individual `get_lock` and `remove_lock` methods?

WRFitch (commented):
Lgtm, I have a couple of points of confusion though

src/mysqlsh_helpers.py
```diff
@@ -311,6 +367,200 @@ def add_instance_to_cluster(self, instance_address, instance_unit_label) -> None
...
+        the lock.
+
+        Args:
+            unit_label: The label for this unit's instance (to be torn down)
+        """
+        try:
+            # Get the cluster primary's address to direct lock acquisition request to.
+            primary_address = self._get_cluster_primary_address()
+            if not primary_address:
+                raise MySQLRemoveInstanceError("Unable to retrieve the cluster primary's address")
```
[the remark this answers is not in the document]
shayancanonical: Added in d85e0fd7a03f66624837d67bc209d913fa533437

WRFitch (approved):

src/mysqlsh_helpers.py
```diff
@@ -57,6 +67,34 @@ class MySQLServiceNotRunningError(Exception):
     pass
 
 
+class MySQLConfirmAllInstancesOnlineError(Exception):
```
marceloneppel, line 70: `is_instance_configured_for_innodb` seems to not use it.

marceloneppel (approved):
Great implementation, Shayan! I really like the approach with the locks and to know about overriding tenacity retry mechanism on tests. I left only two comments.

src/mysqlsh_helpers.py
```diff
@@ -311,6 +381,259 @@ def add_instance_to_cluster(self, instance_address, instance_unit_label) -> None
                     f"Failed to add instance {instance_address} to cluster {self.cluster_name} with recovery method 'auto'. Trying method 'clone'"
                 )
 
+    def is_instance_configured_for_innodb(
+        self, instance_address: str, instance_unit_label: str
+    ) -> bool:
+        """Confirm if instance is configured for use in an InnoDB cluster.
+
+        Args:
+            instance_address: The instance address for which to confirm InnoDB configuration
```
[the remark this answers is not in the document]
shayancanonical: added in 5b30da9c233a0000a12b0c3dfb1a9da7dd6327f2

src/mysqlsh_helpers.py
shayancanonical: great catch! this was added with some other code that is since deleted. removed in 5b30da9c233a0000a12b0c3dfb1a9da7dd6327f2

marceloneppel (approved):

shayancanonical:
Thanks for the reviews!

metadata.yaml
```diff
@@ -9,12 +9,18 @@ summary: |
   MySQL is  a widely used, open-source relational database management system
   (RDBMS). MySQL InnoDB cluster provides a complete high availability solution
   for MySQL via Group Replication.
+source: https://github.com/canonical/mysql-operator
+issues: https://github.com/canonical/mysql-operator/issues
```
taurus-forever, line 13: For the history: https://discourse.charmhub.io/t/question-why-charm-mysql-ignores-issues-attribute-in-metadata-yaml/10029/3

PR #10, opened 2022-05-10
Implement scaling and replication related integration tests

shayancanonical:
# Issue
- Non-leader units are not moved from `WaitingStatus` to `ActiveStatus` when the unit joins a cluster
- We do not have actions to retrieve either `cluster admin` or `server config` credentials
- We do not have integration tests for scaling up+down and replication

# Solution
- Move non-leader units from `WaitingStatus` to `ActiveStatus` in the update-status event handler if the unit is part of the cluster
- Implement actions to retrieve `cluster admin` and `server config` credentials
- Implement integration tests for scaling up+down and replication

# Context
I am using an annotation `pytest.mark.order()` (provided by the `pytest-order` package) to serialize the integration tests in a specific order

# Release Notes
- Implement scaling and replication related integration tests

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -157,6 +157,9 @@ class MySQLOperatorCharm(CharmBase):
 
     def _on_update_status(self, _) -> None:
         """Handle the update_status event."""
+        if not self._is_peer_data_set:
+            return
+
         unit_label = self.unit.name.replace("/", "-")
         if isinstance(self.unit.status, WaitingStatus) and self._mysql.is_instance_in_cluster(
             unit_label
--- a/src/charm.py
+++ b/src/charm.py
@@ -9,7 +9,7 @@ import logging
 import secrets
 import string
 
-from ops.charm import CharmBase, RelationJoinedEvent, StartEvent
+from ops.charm import ActionEvent, CharmBase, RelationJoinedEvent, StartEvent
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus, WaitingStatus
 
@@ -23,6 +23,8 @@ from mysqlsh_helpers import (
 
 logger = logging.getLogger(__name__)
 
+CLUSTER_ADMIN_USERNAME = "clusteradmin"
+SERVER_CONFIG_USERNAME = "serverconfig"
 PASSWORD_LENGTH = 24
 PEER = "database-peers"
 
@@ -63,8 +65,16 @@ class MySQLOperatorCharm(CharmBase):
         self.framework.observe(
             self.on.database_storage_detaching, self._on_database_storage_detaching
         )
+        self.framework.observe(self.on.update_status, self._on_update_status)
 
         self.framework.observe(self.on[PEER].relation_joined, self._on_peer_relation_joined)
+
+        self.framework.observe(
+            self.on.get_cluster_admin_credentials_action, self._on_get_cluster_admin_credentials
+        )
+        self.framework.observe(
+            self.on.get_server_config_credentials_action, self._on_get_server_config_credentials
+        )
 
     # =======================
     #  Charm Lifecycle Hooks
@@ -145,6 +155,14 @@ class MySQLOperatorCharm(CharmBase):
 
         self.unit.status = ActiveStatus()
 
+    def _on_update_status(self, _) -> None:
+        """Handle the update_status event."""
+        unit_label = self.unit.name.replace("/", "-")
+        if isinstance(self.unit.status, WaitingStatus) and self._mysql.is_instance_in_cluster(
+            unit_label
+        ):
+            self.unit.status = ActiveStatus()
+
     def _on_peer_relation_joined(self, event: RelationJoinedEvent) -> None:
         """Handle the peer relation joined event."""
         # Only execute in the unit leader
@@ -173,6 +191,32 @@ class MySQLOperatorCharm(CharmBase):
         self._mysql.remove_instance(unit_label)
 
     # =======================
+    #  Custom Action Handlers
+    # =======================
+
+    def _on_get_cluster_admin_credentials(self, event: ActionEvent) -> None:
+        """Action used to retrieve the cluster admin credentials."""
+        event.set_results(
+            {
+                "cluster-admin-username": CLUSTER_ADMIN_USERNAME,
+                "cluster-admin-password": self._peers.data[self.app].get(
+                    "cluster-admin-password", "<to_be_generated>"
+                ),
+            }
+        )
+
+    def _on_get_server_config_credentials(self, event: ActionEvent) -> None:
+        """Action used to retrieve the server config credentials."""
+        event.set_results(
+            {
+                "server-config-username": SERVER_CONFIG_USERNAME,
+                "server-config-password": self._peers.data[self.app].get(
+                    "server-config-password", "<to_be_generated>"
+                ),
+            }
+        )
+
+    # =======================
     #  Helpers
     # =======================
 
@@ -185,9 +229,9 @@ class MySQLOperatorCharm(CharmBase):
             self.model.get_binding(PEER).network.bind_address,
             peer_data["cluster-name"],
             peer_data["root-password"],
-            "serverconfig",
+            SERVER_CONFIG_USERNAME,
             peer_data["server-config-password"],
-            "clusteradmin",
+            CLUSTER_ADMIN_USERNAME,
             peer_data["cluster-admin-password"],
         )
 
--- a/src/mysqlsh_helpers.py
+++ b/src/mysqlsh_helpers.py
@@ -409,6 +409,33 @@ class MySQL:
             )
             return False
 
+    def is_instance_in_cluster(self, unit_label):
+        """Confirm if instance is in the cluster.
+
+        Args:
+            unit_label: The label of unit to check existence in cluster for
+
+        Returns:
+            Boolean indicating whether the unit is a member of the cluster
+        """
+        commands = (
+            f"shell.connect('{self.cluster_admin_user}:{self.cluster_admin_password}@{self.instance_address}')",
+            f"cluster = dba.get_cluster('{self.cluster_name}')",
+            f"print(cluster.status()['defaultReplicaSet']['topology'].get('{unit_label}', {{}}).get('status', 'NOT_A_MEMBER'))",
+        )
+
+        try:
+            logger.debug(f"Checking existence of unit {unit_label} in cluster {self.cluster_name}")
+
+            output = self._run_mysqlsh_script("\n".join(commands))
+            return "ONLINE" in output
+        except subprocess.CalledProcessError:
+            # confirmation can fail if the clusteradmin user does not yet exist on the instance
+            logger.debug(
+                f"Failed to check existence of unit {unit_label} in cluster {self.cluster_name}"
+            )
+            return False
+
     @retry(
         retry=retry_if_exception_type(MySQLRemoveInstanceRetryError),
         stop=stop_after_attempt(15),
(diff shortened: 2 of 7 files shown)


tests/integration/test_charm.py
```diff
@@ -37,24 +42,232 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+    time.sleep(5)
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME],
+        status="active",
+        raise_on_blocked=True,
+        timeout=1000,
+    )
+
+    # Wait for unit to be destroyed and confirm that the new primary unit is different
+    assert len(ops_test.model.applications[APP_NAME].units) == 2
```
WRFitch, line 123: Though I like the `@pytest.mark.order(x)` flags, you could make the order more flexible by scaling back up to 3 units at the end of this test.

shayancanonical: That's a great suggestion! It's good to assume idempotency event if there's an established order of execution for tests. Addressed in [7972d12](https://github.com/canonical/mysql-operator/pull/10/commits/7972d128c39bbf46d38369d4579cc2808d34c85d)

paulomach (approved):
LGTM. I want to discuss later how to abstract some of the methods for running SQL in units so the our tests can be easily shared across VM/k8s

tests/integration/test_charm.py
```diff
@@ -37,24 +42,248 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+    await ops_test.model.applications[APP_NAME].add_unit()
+    time.sleep(5)
+    await ops_test.model.wait_for_idle(
+        apps=[APP_NAME],
+        status="active",
+        raise_on_blocked=True,
+        timeout=1000,
+    )
+
+    assert len(ops_test.model.applications[APP_NAME].units) == 4
```
WRFitch, line 191: A helper function might clean these lines up a bit: 

```python
def assertHasUnits(expected_units: int): 
    assert len(ops_test.model.applications[APP_NAME].units) == expected_units
```

tests/integration/test_charm.py
```diff
@@ -37,24 +42,248 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+        output = await run_command_on_unit(
+            unit,
+            f"{mysqlsh_sql_base_command} -e \"SELECT * FROM test.instance_state_replication WHERE id = '{random_chars}'\"",
+        )
+        assert random_chars in output
+
+    await ops_test.model.set_config({"update-status-hook-interval": "60m"})
+
+
+async def get_primary_unit(
```
WRFitch, line 230: Can this be moved to tests/integration/helpers.py?

src/charm.py
```diff
@@ -143,6 +159,8 @@ def _on_start(self, event: StartEvent) -> None:
             self.unit.status = BlockedStatus("Failed to initialize juju units operations table")
             return
 
+        self._peers.data[self.app]["units-added-to-cluster"] = "1"
+
```
marcoppenheimer, line 163: I wonder if this is informative. 
Without understanding too much context, would it be useful to have a dict of unit-names within the application data, added when a unit starts, and then whenever we needed this key, a unit could just count keys in the app data?

marcoppenheimer: My thinking being that that way, we could add unit-specific data easily, without adding non-user-helpful information in the relation data.

src/mysqlsh_helpers.py
```diff
@@ -409,6 +409,33 @@ def is_instance_configured_for_innodb(
...
+
+        try:
+            logger.debug(f"Checking existence of unit {unit_label} in cluster {self.cluster_name}")
+
+            output = self._run_mysqlsh_script("\n".join(commands))
+            return "ONLINE" in output
+        except subprocess.CalledProcessError:
+            # confirmation can fail if the clusteradmin user does not yet exist on the instance
+            logger.debug(
+                f"Failed to check existence of unit {unit_label} in cluster {self.cluster_name}"
```
marcoppenheimer, line 435: Does this fail to 'check' existence, or 'confirm' existence?

tests/integration/test_charm.py
```diff
@@ -37,24 +42,248 @@ async def test_build_and_deploy(ops_test: OpsTest):
         raise_on_blocked=True,
         timeout=1000,
     )
-    assert len(ops_test.model.applications[APP_NAME].units) == 1
-    assert ops_test.model.applications[APP_NAME].units[0].workload_status == "active"
+    assert len(ops_test.model.applications[APP_NAME].units) == 3
+
+    for unit in ops_test.model.applications[APP_NAME].units:
+        assert unit.workload_status == "active"
```
marcoppenheimer, line 48: I love this simple test.

tests/integration/test_charm.py
```diff
@@ -37,24 +42,248 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+
+    primary_unit = await get_primary_unit(
+        ops_test,
+        random_unit,
+        server_config_credentials["username"],
+        server_config_credentials["password"],
+    )
+    primary_unit_name = primary_unit.name
+
+    assert len(ops_test.model.applications[APP_NAME].units) == 3
```
marcoppenheimer, line 111: Do you need to re-assert this if it's passed above? I probably missed something. As in, these tests run within the same application no?

tests/integration/test_charm.py
```diff
@@ -37,24 +42,248 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+        "CREATE DATABASE IF NOT EXISTS test",
+        "CREATE TABLE IF NOT EXISTS test.instance_state_replication (id varchar(40), primary key(id))",
+        f"INSERT INTO test.instance_state_replication VALUES ('{random_chars}')",
+    )
+
+    mysqlsh_sql_base_command = f"mysqlsh --sql {server_config_credentials['username']}:{server_config_credentials['password']}@127.0.0.1"
+
+    await run_command_on_unit(
+        primary_unit, f"{mysqlsh_sql_base_command} -e \"{';'.join(create_records_sql)}\""
+    )
```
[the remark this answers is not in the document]
marcoppenheimer: I think it's fine to duplicate this, too much abstraction in tests is more of a headache.

tests/integration/test_charm.py
marcoppenheimer, line 191: This is an opinion, and applies to other tests above. 
Are these `assert`ions part of the test, or are they just 'if' statements?

If the test fails because there weren't 4 units for some unknown reason, it doesn't mean that `cluster_preserves_data_on_delete` wasn't met, which would make debugging more complicated. Maybe make these if-statements, and fail the test with a helpful message. Same to the tests above. We should probably keep assertions to failure conditions for the 'thing' we're testing, not for validating set-up.

Again, it's possible I am misunderstanding.

tests/integration/test_charm.py
```diff
@@ -37,24 +42,248 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+    if not matches:
+        return None
+
+    cluster_status = json.loads(matches.group(1).strip())
+
+    primary_name = [
+        label
+        for label, member in cluster_status["defaultReplicaSet"]["topology"].items()
+        if member["mode"] == "R/W"
+    ][0].replace("-", "/")
```
marcoppenheimer, line 265: Might be more readable as a bog-standard `for` loop, `black` formatting for list-comprehension is a pain.

tests/integration/test_charm.py
```diff
@@ -37,24 +42,248 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+    )
+
+    random_chars = generate_random_string(40)
+    create_records_sql = (
+        "CREATE DATABASE IF NOT EXISTS test",
+        "CREATE TABLE IF NOT EXISTS test.data_replication_table (id varchar(40), primary key(id))",
+        f"INSERT INTO test.data_replication_table VALUES ('{random_chars}')",
+    )
+
+    mysqlsh_sql_base_command = f"mysqlsh --sql {server_config_credentials['username']}:{server_config_credentials['password']}@127.0.0.1"
```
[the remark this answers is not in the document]
shayancanonical: paulo suggested that we use a mysql client, which i have added in ff877d16df482a9781067a33b0a1cf1e46591782 and refactored this out

tests/integration/test_charm.py
```diff
@@ -37,24 +42,248 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+
+    await run_command_on_unit(
+        primary_unit, f"{mysqlsh_sql_base_command} -e \"{';'.join(create_records_sql)}\""
```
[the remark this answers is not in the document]
shayancanonical: was refactored out in ff877d16df482a9781067a33b0a1cf1e46591782

tests/integration/test_charm.py
```diff
@@ -37,24 +42,248 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+        random_unit,
+        server_config_credentials["username"],
+        server_config_credentials["password"],
+    )
+
+    assert primary_unit_name != new_primary_unit.name
+
+    # Add the unit back and wait until it is active
+    await ops_test.model.applications[APP_NAME].add_unit()
+    time.sleep(5)
```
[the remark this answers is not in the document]
shayancanonical: im now actually using `block_until` to wait until N units are available, and then using `wait_for_idle` for the added unit to reach ActiveStatus

tests/integration/test_charm.py
```diff
@@ -37,24 +42,248 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+        server_config_credentials["username"],
+        server_config_credentials["password"],
+    )
+
+    random_chars = generate_random_string(40)
+    create_records_sql = (
+        "CREATE DATABASE IF NOT EXISTS test",
+        "CREATE TABLE IF NOT EXISTS test.data_replication_table (id varchar(40), primary key(id))",
+        f"INSERT INTO test.data_replication_table VALUES ('{random_chars}')",
+    )
```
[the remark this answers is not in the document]
shayancanonical: I agree with Marc, it seems a bit too much to abstract this. Keeping it inline and verbose allows us to easily modify the sequence of commands being executed if need be in the future. Also, it is more readable like this

tests/integration/test_charm.py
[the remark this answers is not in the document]
shayancanonical: I agree, still leaning towards duplicating as it is more readable

tests/integration/test_charm.py
shayancanonical: deleted these assert statements in ff877d16df482a9781067a33b0a1cf1e46591782 (as Marc suggested that they're redundant)

tests/integration/test_charm.py
shayancanonical: moved in ff877d16df482a9781067a33b0a1cf1e46591782

src/charm.py
shayancanonical: So the reason we are setting data in the peer relation databag is to trigger `relation_changed` events on peers. This is where peers (non-leader) move from `WaitingStatus` (waiting to join cluster) to `ActiveStatus`. I deliberately used a counter instead of a list of unit names to avoid implementing a teardown hook just to remove data from the peer relation databag

tests/integration/test_charm.py
shayancanonical: Removed assertions in ff877d16df482a9781067a33b0a1cf1e46591782 (where I am only testing the 'thing' to be tested)

tests/integration/test_charm.py
shayancanonical: Refactored to be a one liner in ff877d16df482a9781067a33b0a1cf1e46591782

WRFitch (approved):

src/mysqlsh_helpers.py
shayancanonical: Renamed to `confirm` existence in 8f190c92787c136efedeb83716867e0363380257

PR #13, opened 2022-06-13
Add support to `shared-db` legacy relation

paulomach:
# Issue
Support legacy relation `shared-db`  (2/2). Previous PR #12 already implemented the `db-router` (1/2).

# Solution

The solution implemented the legacy relationship by emulating its behavior, as observed in practice with the handy help of [jhack's](https://github.com/PietroPasotti/jhack) `show relation` feature.

The behavior is:
1. When legacy relation is added, create database and user for given username and database name present in relation data.
2. Populate credentials and database endpoint info in the relation data.
3. Remove user (preserve database) when relation is broken

# Testing

* Test was done by extending existing unit tests.
* Integrations tests to come in subsequent PR's (we are in discussion on the best way to add tests for legacy relations). 
* Manual validations with OSM and Keystone charms

# Release Notes

* `lib/charms/mysql/v0/mysql.py` Add user removal function and companion custom error
* `src/charm.py` Add handlers for `shared-db` relation changed and broken
* `tests/unit/test_charm.py` Unit test for new charm handlers
* `tests/unit/test_mysql.py` Unit test for new lib function
* `metadata.yaml` Register relation


src/charm.py
```diff
@@ -291,6 +303,98 @@ def _on_db_router_relation_changed(self, event: RelationChangedEvent) -> None:
...
+            unit_relation_databag["db_host"] = cluster_primary
+            # Database port is static in legacy charm
+            unit_relation_databag["db_port"] = "3306"
+            # Wait timeout is a config option in legacy charm
+            # defaulted to 3600 seconds
+            unit_relation_databag["wait_timeout"] = "3600"
+            unit_relation_databag["password"] = password
+
+            unit_names = " ".join([unit.name for unit in event.relation.units])
+            unit_relation_databag["allowed_units"] = unit_names
```
shayancanonical, line 360: im wondering if we should be updating `allowed_units` with every unit that joins the relation (and not only the first time the database user and application database are created)

shayancanonical: now that i think of it, we may also need to update `allowed_units` upon the relation departed event (when units leave the relation)

src/charm.py
```diff
@@ -291,6 +303,98 @@ def _on_db_router_relation_changed(self, event: RelationChangedEvent) -> None:
...
+        """
+        if not self.unit.is_leader():
+            return
+
+        app_relation_databag = event.relation.data[self.app]
+        username = app_relation_databag.get(f"relation_id_{event.relation.id}_db_user")
+
+        if not username:
+            # Can't do much if we don't have the username
+            return
```
WRFitch, line 388: Should this be logged in some way? If there's no username where we expect one to be, it might be a bad idea to fail quietly.

src/charm.py
```diff
@@ -291,6 +303,98 @@ def _on_db_router_relation_changed(self, event: RelationChangedEvent) -> None:
...
+        requires_relation_databag = event.relation.data[event.unit]
+        database_name = requires_relation_databag.get("database")
+        database_user = requires_relation_databag.get("username")
+
+        if not database_name or not database_user:
+            # Cannot create scoped database without credentials
+            logger.warning(
+                "Missing information for shared-db relation to create a database and scoped user"
+            )
+            self.unit.status = ActiveStatus()
```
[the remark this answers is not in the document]
paulomach: I'm never sure. My though here is, in a scenario of a running cluster with multiple apps, one should not block/wait a unit when the error is most probably come from the consumer charm.
In that manner, the unit just resume it's active status since it was previously put in Maintenance.

src/charm.py
```diff
@@ -291,6 +303,98 @@ def _on_db_router_relation_changed(self, event: RelationChangedEvent) -> None:
...
+
+        password = generate_random_password(PASSWORD_LENGTH)
+
+        try:
+            self._mysql.create_application_database_and_scoped_user(
+                database_name, database_user, password
+            )
+
+            # set the relation data for consumption
+            cluster_primary = str(self.model.get_binding(PEER).network.bind_address)
```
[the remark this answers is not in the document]
paulomach: That's a way. I also wonder if we should ensure that the leader unit matches cluster primary (as the reactive charm does).

src/charm.py
paulomach: You are right, but bear in mind this is mainly a backward compatibility thing, since this data is not being used. IIRC the reactive charm did not allow connections from every host (e.g. `bind 0.0.0.0`). Since it's not the case here, this does not really matter - will add the relation departing part though.

src/charm.py
```diff
@@ -291,6 +303,98 @@ def _on_db_router_relation_changed(self, event: RelationChangedEvent) -> None:
...
+
+            unit_names = " ".join([unit.name for unit in event.relation.units])
+            unit_relation_databag["allowed_units"] = unit_names
+
+            # store username for relation in app databag
+            # this is used to remove the user when the relation is broken
+            app_relation_databag[f"relation_id_{event.relation.id}_db_user"] = database_user
+
+        except (
+            MySQLClientError,
```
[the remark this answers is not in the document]
paulomach: It wont - good catch.

jnsgruk:
(This is just a suggestion, and you're very free to ignore, it just occurred to me reading through)

Is it worth breaking out each of the legacy relations into their own file (similar to a library, but not quite).

Example: ./src/relations/shared-db.py

This would allow us to keep the code for each relation separate, possibly aiding debugging and testing? Could take a similar for to a traditional relation lib where you pass and instance of the charm into the constructor, which then registers events, etc.

Are we planning on providing consumer libs for the legacy relations, or concentrating the efforts on the new relations there and just supporting legacy relations for those charms that already have them implemented?

(cc @marceloneppel @delgod @paulomach @shayancanonical)

WRFitch (approved):
LGTM! Also I like Jon's suggestion, we're already splitting out the shared postgres `db` and `db-admin` relations into a charm library, so having the `backend-db-admin` relation integrated into the pgb charm is awkward.

config.yaml
```diff
@@ -5,3 +5,9 @@ options:
   cluster-name:
     description: "Optional - Name of the MySQL InnoDB cluster"
     type: "string"
+  auto-delete:
+    type: boolean
+    description: |
+        When a relation is removed, auto-delete ensures that any relevant databases
+        associated with the relation are also removed
+    default: false
```
shayancanonical, line 13: missing new line

src/charm.py
```diff
@@ -291,6 +308,131 @@ def _on_db_router_relation_changed(self, event: RelationChangedEvent) -> None:
...
+            # set the relation data for consumption
+            cluster_primary = self._mysql.get_cluster_primary_address()
+
+            unit_relation_databag["db_host"] = cluster_primary.split(":")[0]
+            # Database port is static in legacy charm
+            unit_relation_databag["db_port"] = "3306"
+            # Wait timeout is a config option in legacy charm
+            # defaulted to 3600 seconds
+            unit_relation_databag["wait_timeout"] = "3600"
+            unit_relation_databag["password"] = password
```
shayancanonical, line 362: soemthing to confirm using jhack - are the values in the databag json encoded? they were for `db-router` (strings were `"string_value"` instead of `string_value`)

shayancanonical (approved):
LGTM, just left one comment

src/charm.py
paulomach: I've noticed that while investigating, and `shared-db` does not json encodes it - so I choose to keep it *fac-simile* from the original one.

PR #15, opened 2022-06-23
Refactor legacy relation code into library like files

shayancanonical:
# Issue
1. There are some bugs preventing the `db-router` relation from working as it is on the `main` branch. Additionally, there is some stale code related to deleting databases upon legacy relation broken (it is stale because it is impossible to tell if the database is still required by another `db-router` or `shared-db` relation)
2. The `db-router` relation users are created with `%` hostnames, whereas they should be created with hostnames equivalent to the consumer unit's private ip address
3. The `db-router` and `shared-db` related code is interleaved with the charm code
4. There is no cleanup of users when the `db-router` consumer application unit departs.

# Solution
1. Fix the bugs, and remove the stale code
2. Correctly create users with hostnames for the legacy relations
3. Break out the `db-router` related code into a separate lib-like file (and move the `db-router` related tests into a separate file). Delete the `shared-db` related code and unittests (they will be added in a lib-like file in a separate PR)
4. Use MySQL user attributes to tag users created for an application unit, and correctly delete them by attribute when the remote unit departs

# Testing
Integration tests are yet to be implement (incoming in a different commit or PR)

# Release Notes
Refactor legacy relation code into separate library like files and make the implementation more robust (with respect to working correctly with the legacy consumer applications)


lib/charms/mysql/v0/mysql.py
```diff
@@ -321,68 +322,53 @@ def configure_mysqlrouter_user(self, username: str, password: str) -> None:
...
-        Args:
-            username: The username for the mysql user
-            hostname (optional): The hostname for the mysql user (defaults to %)
-
-        Raises MySQLRemoveUserError
-            if there is an issue removing the mysql user
-        """
-        remove_user_commands = (f"DROP USER IF EXISTS '{username}'@'{hostname}'",)
+    def delete_users_for_unit(self, unit_name: str) -> None:
+        """Delete users for a unit."""
```
paulomach, line 358: add `Args:` in docstring

paulomach (changes requested):
Just a thing to check `on_leader_elected`

src/relations/db_router.py
```diff
@@ -0,0 +1,254 @@
...
+        Copy data from the relation's application databag to the leader unit databag
+        since legacy applications expect credential data to be populated on the leader
+        unit databag.
+        """
+        relation = self.charm.model.get_relation(LEGACY_DB_ROUTER)
+        if not relation:
+            return
+
+        relation_databag = relation.data
+        for key, value in relation_databag[self.charm.app].items():
```
paulomach, line 166: Just got another case: when syncing, it's possible that the departed unit was db primary (`db_host`), so this needs to be updated also.

shayancanonical: Good catch - I was setting the `db_host` as the leader unit's address rather than the cluster primary. Addressed in [b4f569d](https://github.com/canonical/mysql-operator/pull/15/commits/b4f569df9c33bec957c9f3eafd520c9950054b00)

src/relations/db_router.py
```diff
@@ -0,0 +1,254 @@
...
+        return user_passwords, requested_user_applications
+
+    def _on_leader_elected(self, _) -> None:
+        """Handle the leader elected event.
+
+        Copy data from the relation's application databag to the leader unit databag
+        since legacy applications expect credential data to be populated on the leader
+        unit databag.
+        """
+        relation = self.charm.model.get_relation(LEGACY_DB_ROUTER)
```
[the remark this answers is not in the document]
shayancanonical: Great catch! Addressed in [b4f569d](https://github.com/canonical/mysql-operator/pull/15/commits/b4f569df9c33bec957c9f3eafd520c9950054b00)

lib/charms/mysql/v0/mysql.py
shayancanonical: Addressed in [b4f569d](https://github.com/canonical/mysql-operator/pull/15/commits/b4f569df9c33bec957c9f3eafd520c9950054b00)

marceloneppel (approved):
LGTM! I really liked the integration tests on the relation.

shayancanonical:
Thanks for the reviews!

PR #16, opened 2022-06-28
Legacy shared_db refactor

paulomach:
# Issue

Implementation of `shared-db` legacy interface was not addressing all edge cases.
Also `charm.py` was starting to become seriously bloated with legacy relations specific code.

# Solution

Robust and feature complete implementation for `shared-db` legacy relation.
Implemented in library like fashion, i.e. decoupled from `charm.py` file.

# Context

`shared-db` interface offers a way for direct accessing mysql cluster, by exposing credentials and primary address.
Each application demands a database name, which is created if not exist.
Each unit related through this interface has an specific user created for it, and access granted to the created database. User password is cached on peer relation and reused across applications that demand the same username. The password must survive relation breakage.
The relation data is consumed by the app through the leader unit databag of mysql (instead of the app databag), which is due app databag not being implemented for legacy charms. Relation data is replicated on the app databag and updated from the app databag for leader reelection scenarios.

# Testing

Testing is implemented in unit and integrations tests.
Integration test make use of the [keystone](https://charmhub.io/keystone) charm.
**NOTE** Unit is currently failing due a bug in `ops.testing` as waiting [this PR](https://github.com/canonical/operator/pull/790) to be merged

# Release Notes

* M [requirements.txt](https://github.com/canonical/mysql-operator/compare/feature/shared_db_refactor?expand=1#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552) - bump version to use `ops_test.fast_forward`
* M [src/charm.py](https://github.com/canonical/mysql-operator/compare/feature/shared_db_refactor?expand=1#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307c) - Instantiate `shared_db`
* A [src/relations/shared_db.py](https://github.com/canonical/mysql-operator/compare/feature/shared_db_refactor?expand=1#diff-2eed71cdafb7ba6e142b4fe80c98d5235f1a61c75b2155ca7d547ccf72210412) - `shared_db` implementation
* A [tests/integration/test_shared_db.py](https://github.com/canonical/mysql-operator/compare/feature/shared_db_refactor?expand=1#diff-6b556c258a790337fe4a26216f7cb5f5613b44565e90a743319243702c9e0f86) - itests with keystone charm
* A [tests/unit/test_shared_db.py](https://github.com/canonical/mysql-operator/compare/feature/shared_db_refactor?expand=1#diff-067a4bf78ac4fb092a24a5ecec3157d2e65b179105ce5373c43cf56d3d7627e6) - utests

### DIFF
--- a/tests/unit/test_shared_db.py
+++ b/tests/unit/test_shared_db.py
@@ -2,9 +2,10 @@
 # See LICENSE file for licensing details.
 
 import unittest
-from unittest.mock import patch
+from unittest.mock import PropertyMock, patch
 
 from charms.mysql.v0.mysql import MySQLCreateApplicationDatabaseAndScopedUserError
+from ops.charm import RelationDepartedEvent
 from ops.model import BlockedStatus
 from ops.testing import Harness
 
@@ -132,6 +133,14 @@ class TestSharedDBRelation(unittest.TestCase):
                 "private-address": "1.1.1.2",
             },
         )
-        self.harness.remove_relation_unit(self.shared_db_relation_id, "other-app/0")
+
+        # Workaround ops.testing not setting `departing_unit` in v1.5.0
+        # ref https://github.com/canonical/operator/pull/790
+        shared_db_relation = self.charm.model.get_relation(LEGACY_DB_SHARED)
+        with patch.object(
+            RelationDepartedEvent, "departing_unit", new_callable=PropertyMock
+        ) as mock_departing_unit:
+            mock_departing_unit.return_value = list(shared_db_relation.units)[0]
+            self.harness.remove_relation_unit(self.shared_db_relation_id, "other-app/0")
 
         _delete_users_for_unit.assert_called_once_with("other-app/0")


src/relations/shared_db.py
```diff
@@ -0,0 +1,182 @@
...
+        )
+        self.framework.observe(
+            self._charm.on[LEGACY_DB_SHARED].relation_departed, self._on_shared_db_departed
+        )
+
+        self.framework.observe(self._charm.on.leader_elected, self._on_leader_elected)
+
+    @property
+    def _peers(self):
+        return self.model.get_relation(PEER)
```
shayancanonical, line 39: self does not have a `model`. did you mean `self.charm.model.get_relation(PEER)`? in fact, i dont think this is being used

shayancanonical (approved):
LGTM, just a minor comment about an unused property

src/relations/shared_db.py
paulomach: Actually that's a neat thing. Because `SharedDBRelation` inherits from `ops.framework.Object` it has direct access to `model`. The `_ peers` is being used for caching user password.

tests/integration/test_shared_db.py
```diff
@@ -0,0 +1,253 @@
...
+        for unit in ops_test.model.applications[KEYSTONE_APP_NAME].units:
+            unit_address = await unit.get_public_address()
+
+            keystone_users.append(f"keystone@{unit_address}")
+
+        await check_keystone_users_existence(
+            ops_test, server_config_credentials, keystone_users, []
+        )
+
+        # Deploy and test another deployment of keystone
```
WRFitch, line 194: This test is pretty big. It might be useful to split it into a few tests, so we can easily pinpoint what functionality causes it to fail.

tests/integration/test_shared_db.py
```diff
@@ -0,0 +1,253 @@
...
+        )
+
+        # Scale mysql back up to 3 units
+        await scale_application(ops_test, APP_NAME, 3)
+
+        # Scale down the first deployment of keystone
+        await scale_application(ops_test, KEYSTONE_APP_NAME, 0)
+        await ops_test.model.remove_application(KEYSTONE_APP_NAME, block_until_done=True)
+
+        # Scale down the mysql application
```
WRFitch, line 251: ops-test creates a separate model for each test file, which is deleted after the run, so you don't have to run these cleanup steps unless you're planning to add more tests later on.

tests/unit/test_shared_db.py
```diff
@@ -0,0 +1,139 @@
...
+            {
+                "database": "shared_database",
+                "hostname": "1.1.1.2",
+                "username": "shared_user",
+            },
+        )
+
+        self.assertTrue(isinstance(self.harness.model.unit.status, BlockedStatus))
+
+        _create_application_database_and_scoped_user.reset_mock()
```
WRFitch, line 108: I don't understand the purpose of resetting the mock here, surely it's out of scope after this test finishes?

WRFitch (approved):
LGTM, I have a couple of questions about the integration testing but nothing blocking.

src/relations/shared_db.py
```diff
@@ -0,0 +1,182 @@
...
+
+        if joined_unit in local_unit_data.get("allowed_units", ""):
+            # Test if relation data is already set for the unit
+            # and avoid re-running it
+            logger.warning(f"Unit {joined_unit} already added to relation")
+            return
+
+        # retrieve data from the relation databag
+        # Abstracted is the fact that it is received from the leader unit of the related app
+        # fallback to peer data if data is not set for the unit (non leader unit)
```
marceloneppel, line 107: I was a little confused by this comment about `fallback to peer data`. Shouldn't it be `fallback to local application data`?

marceloneppel (approved):
LGTM! I have just one question about one comment in the code.

tests/unit/test_shared_db.py
paulomach: sloppy C-c, C-v - good catch

tests/integration/test_shared_db.py
paulomach: This is useful to catch any error in the departing of units - there's some activity in the relation management tied to it.

tests/integration/test_shared_db.py
paulomach: Indeed is huge - will evaluate what I can do about it

tests/integration/test_shared_db.py
WRFitch: Ah, I didn't think of that!

PR #19, opened 2022-07-06
Direct writes to the cluster primary for methods in mysql lib

shayancanonical:
# Issue
1. For some methods in the `mysql` lib (`configure_mysqlrouter_user` and `create_application_database_and_scoped_user`), the code assumes that the juju leader unit is also the cluster primary. If it is not, the writes for the methods will fail.
2. After the `root` user is created, subsequent usages of `run_mysqlcli_script` need to pass in the `root` user's password

# Solution
1. Refactor the two methods to direct the writes to the cluster primary.
2. Pass in the `root` user password in `does_mysql_user_exist` method

# Release Notes
Fix bug and direct writes to the cluster primary for methods in the mysql lib

### DIFF
--- a/tests/unit/test_mysql.py
+++ b/tests/unit/test_mysql.py
@@ -94,7 +94,9 @@ class TestMySQLBase(unittest.TestCase):
         )
 
         self.mysql.does_mysql_user_exist("test_username", "1.1.1.1")
-        _run_mysqlcli_script.assert_called_once_with("\n".join(user_existence_command))
+        _run_mysqlcli_script.assert_called_once_with(
+            "\n".join(user_existence_command), password="password"
+        )
 
         # Reset the mock
         _run_mysqlcli_script.reset_mock()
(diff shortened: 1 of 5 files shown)


tests/unit/test_mysql.py
```diff
@@ -115,34 +119,38 @@ def test_does_mysql_user_exist_failure(self, _run_mysqlcli_script):
         with self.assertRaises(MySQLCheckUserExistenceError):
             self.mysql.does_mysql_user_exist("test_username", "1.1.1.1")
 
-    @patch("charms.mysql.v0.mysql.MySQLBase._run_mysqlcli_script")
-    def test_configure_mysqlrouter_user(self, _run_mysqlcli_script):
+    @patch("charms.mysql.v0.mysql.MySQLBase.get_cluster_primary_address")
+    @patch("charms.mysql.v0.mysql.MySQLBase._run_mysqlsh_script")
```
zmraul, line 123: Since you are _always_ going to need to patch `_run_mysqlsh_script`, it could be made a pytest fixture with `autouse=True`

zmraul (approved):
Looks good :) I left a suggestion to consider on unit tests

marceloneppel (approved):
LGTM

Issue #23, opened 2022-07-20
"storage detaching" hook fails


barrettj12:
### Bug Description

When trying to remove mysql (revision cs:67), it ends up with status "error" and message `hook failed: "storage-detaching"`.

### To Reproduce

```
juju add-model m
juju deploy cs:mysql-67
juju destroy-model -y m
```

### Environment

Running Juju locally, v3.0-beta3 (SHA fcf07484)
OS: Ubuntu 20.04 LTS
on cloud LXD v5.3-91e042b
using mysql revision 67 (channel candidate)

### Relevant log output

```shell
unit-mysql-0: 11:57:21 ERROR juju.worker.uniter.operation hook "database-storage-detaching" (via hook dispatching script: dispatch) failed: exit status 1
unit-mysql-0: 11:57:21 INFO juju.worker.uniter awaiting error resolution for "storage-detaching" hook
unit-mysql-0: 11:58:01 INFO juju.worker.uniter awaiting error resolution for "storage-detaching" hook
unit-mysql-0: 11:58:02 ERROR unit.mysql/0.juju-log Uncaught exception while in charm code:
Traceback (most recent call last):
  File "./src/charm.py", line 279, in <module>
    main(MySQLOperatorCharm)
  File "/var/lib/juju/agents/unit-mysql-0/charm/venv/ops/main.py", line 431, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-mysql-0/charm/venv/ops/main.py", line 142, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-mysql-0/charm/venv/ops/framework.py", line 316, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-mysql-0/charm/venv/ops/framework.py", line 784, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-mysql-0/charm/venv/ops/framework.py", line 857, in _reemit
    custom_handler(event)
  File "./src/charm.py", line 199, in _on_database_storage_detaching
    self._mysql.remove_instance(unit_label)
  File "./src/charm.py", line 253, in _mysql
    peer_data["cluster-name"],
  File "/var/lib/juju/agents/unit-mysql-0/charm/venv/ops/model.py", line 539, in __getitem__
    return self._data[key]
KeyError: 'cluster-name'
[...]

PR #24, opened 2022-07-20
Add guards to not remove instance from cluster if it is not already a member

shayancanonical:
# Issue
[Issue](https://github.com/canonical/mysql-operator/issues/23)
The above issue is caused by a race condition where the model (and thus cluster) is torn down before being fully initialized. The error trace is encountered when the instance/unit being torn down is not a member of the cluster and is being attempted to be removed from the cluster in the `storage_detached` event handler.

# Solution
Add guards to not attempt to remove the instance from the cluster if it is not already a member of the cluster.

# Release Notes
Add guards to not remove instance from cluster if it is not already a member

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -193,9 +193,18 @@ class MySQLOperatorCharm(CharmBase):
 
     def _on_database_storage_detaching(self, _) -> None:
         """Handle the database storage detaching event."""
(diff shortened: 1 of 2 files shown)


paulomach (approved):
legit

marceloneppel (approved):
LGTM!

PR #28, opened 2022-08-05
MariaDB (aka `mysql`) relation port from k8s charm

paulomach:
# Issue

There's prospective commercial interest on using MySQL with MLFlow in a cross model setting, i.e. MySQL@machine <--> MLFlow@k8s.

# Solution

MLFlow charm uses legacy mariadb relation, previously implemented on k8s.
This charm ports the relation from k8s and implement custom tests for it (since there's no machine charm using this relation).
The test charm is the same used for testing "new" relations, extended for this use case.

# Testing

Testing was implemented in unit and integration test for this relation.
The implementation was also validated manually with the cross model setup with MLFlow.

# Release Notes
It looks a lot, but there's a lot of boiler plate. 

- `.github/workflows/ci.yaml` - added test for relation
- `.github/workflows/release.yaml` - updated release action
- `README.md` - added notice about GR requirement and relation doc
- `config.yaml` - config necessary for this relation
- `lib/charms/mysql/v0/mysql.py` - sanitize user name for safe usage in mysql
- `metadata.yaml` - add relation to metadata
- `pyproject.toml` - new test register
- `src/charm.py` - relation instantiation
- `src/constants.py` - add relation name
- `src/relations/mysql.py` - relation implementation
- `tests/integration/application-charm/lib/charms/data_platform_libs/v0/database_requires.py` - IGNORE - bumped dp-libs
- `tests/integration/application-charm/metadata.yaml` - mariadb relation to test charm
- `tests/integration/application-charm/src/charm.py` - mariadb relation support into the test charm
- `tests/integration/test_database.py` - opportunistic speed improvement on non related test
- `tests/integration/test_relation_mysql_legacy.py` - int test for mariadb relation
- `tests/unit/test_relation_mysql_legacy.py` - unit test for mariadb relation
- `tests/unit/test_mysql.py` - test fix after lib fix
- `tox.ini` - add tox env for int test


src/relations/mysql.py
```diff
@@ -0,0 +1,146 @@
...
+        being elected).
+        """
+        if not self.charm.unit.is_leader():
+            return
+
+        # Wait until on-config-changed event is executed
+        # (wait for root password to have been set)
+        if not self.charm._is_peer_data_set:
+            event.defer()
+            return
```
MiaAltieri, line 95: it is possible that this event will defer continuously and block the `on-config-changed` event from executing, or not block it but lead to a long time for that hook to actually get fired. This will lead to poor performance. Perhaps it would be best to:

```
    def _on_mysql_relation_created(self, event: RelationCreatedEvent) -> None:
        ....
        
        if not self.charm._is_peer_data_set:
            self.app_data['needs relation created']=True
```
and then 
```
     def config_changed -> None:
        ....
        
        if self.app_data['needs relation created']:
            <emit event or call this function>
           self.app_data['needs relation created'] = False
```

src/relations/mysql.py
```diff
@@ -0,0 +1,146 @@
...
+
+        logger.warning("DEPRECATION WARNING - `mysql` is a legacy interface")
+
+        username = self.charm.config.get("mysql-interface-user")
+        database = self.charm.config.get("mysql-interface-database")
+
+        if not username or not database:
```
MiaAltieri, line 102: comment please

src/relations/mysql.py
```diff
@@ -0,0 +1,146 @@
...
+            event.defer()
+            return
+
+        # Only execute if the application user does not exist
```
MiaAltieri, line 106: change to:
`executing when sql user already exists can lead to <explanation>`
your SQL specific knowledge would be super useful here

src/relations/mysql.py
```diff
@@ -0,0 +1,146 @@
...
+        if self.charm._mysql.does_mysql_user_exist(username, "%"):
+            return
+
+        password = self._get_or_set_password_in_peer_databag(username)
+
+        self.charm._mysql.create_application_database_and_scoped_user(
```
MiaAltieri, line 112: do you need to catch `MySQLCreateApplicationDatabaseAndScopedUserError` ? Or is it better to error out here? If its better to error out ignore me :)

src/relations/mysql.py
```diff
@@ -0,0 +1,146 @@
...
+
+        Delete the application user created in the relation created
+        event handler.
+        """
+        if not self.charm.unit.is_leader():
+            return
+
+        logger.warning("DEPRECATION WARNING - `mysql` is a legacy interface")
+
+        self.charm._mysql.delete_users_for_unit("mysql-legacy-relation")
```
MiaAltieri, line 146: same here `MySQLDeleteUsersForUnitError `, should we defer or retry or maybe go into blocked status?

tests/integration/test_mariadb.py
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env python3
```
MiaAltieri, line 1: rename file `tests/integration/test_relation_mariadb.py`

tests/integration/test_mariadb.py
```diff
@@ -0,0 +1,129 @@
...
+
+    async with ops_test.fast_forward():
+        await asyncio.gather(
+            ops_test.model.wait_for_idle(
+                apps=[DATABASE_APP_NAME], status="active", raise_on_blocked=True
+            ),
+            ops_test.model.wait_for_idle(
+                apps=[APPLICATION_APP_NAME], status="waiting", raise_on_blocked=True
+            ),
+        )
```
MiaAltieri, line 129: Add check in this test that verifies loss of acess to DB

tests/integration/test_mariadb.py
```diff
@@ -0,0 +1,129 @@
...
+
+    await ops_test.model.relate(APPLICATION_APP_NAME, f"{DATABASE_APP_NAME}:{ENDPOINT}")
+
+    async with ops_test.fast_forward():
+        await ops_test.model.block_until(
+            lambda: is_relation_joined(ops_test, ENDPOINT, ENDPOINT) == True  # noqa: E712
+        )
+
+        await ops_test.model.wait_for_idle(apps=APPS, status="active")
+
```
MiaAltieri, line 106: add tests for:
1. CRUD ops
2. Admin DB access
3. Relation data
4. Scale MySQL by 3 units, remove the original 3 units, verify app still has access

tests/unit/test_mariadb.py
```diff
@@ -0,0 +1,105 @@
+# Copyright 2022 Canonical Ltd.
```
MiaAltieri, line 1: change file to `test_relation_mariadb.py`

tests/unit/test_mariadb.py
```diff
@@ -0,0 +1,105 @@
...
+
+class TestMariaDBRelation(unittest.TestCase):
+    def setUp(self):
+        self.harness = Harness(MySQLOperatorCharm)
+        self.addCleanup(self.harness.cleanup)
+        self.harness.begin()
+        self.peer_relation_id = self.harness.add_relation("database-peers", "database-peers")
+        self.harness.add_relation_unit(self.peer_relation_id, "mysql/1")
+        # self.maria_db_relation_id = self.harness.add_relation(LEGACY_MYSQL, "other-app")
+        # self.harness.add_relation_unit(self.maria_db_relation_id, "other-app/0")
```
MiaAltieri, line 23: remove

MiaAltieri (changes requested):
Looks awesome 🤩! Some nits, suggestions, and requests

.github/workflows/ci.yaml
```diff
@@ -71,3 +71,15 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Setup operator environment
+        uses: charmed-kubernetes/actions-operator@main
+        with:
+          provider: lxd
+      - name: Run integration database tests
+        run: tox -e integration-mariadb
```
shayancanonical, line 85: recommend renaming `integration-mariadb` -> `integration-myql-interface` to avoid any confusion

tests/integration/application-charm/src/charm.py
```diff
@@ -17,14 +17,15 @@
...
-from ops.charm import CharmBase, RelationChangedEvent
+from ops.charm import CharmBase, RelationChangedEvent, RelationJoinedEvent
 from ops.main import main
 from ops.model import ActiveStatus, WaitingStatus
 
 logger = logging.getLogger(__name__)
 
 PEER = "application-peers"
 REMOTE = "database"
+MARIADB = "mysql"
```
shayancanonical, line 28: rename `mariadb` -> `legacy_mysql`?

tests/integration/test_mariadb.py
```diff
@@ -0,0 +1,129 @@
...
+@pytest.mark.abort_on_fail
+@pytest.mark.mariadb_tests
+async def test_relation_broken(ops_test: OpsTest):
+    """Remove relation and wait for the expected changes in status."""
+    await ops_test.model.applications[DATABASE_APP_NAME].remove_relation(
+        f"{APPLICATION_APP_NAME}:{ENDPOINT}", f"{DATABASE_APP_NAME}:{ENDPOINT}"
+    )
+
+    await ops_test.model.block_until(
+        lambda: is_relation_broken(ops_test, ENDPOINT, ENDPOINT) == True  # noqa: E712
```
shayancanonical, line 118: why the `noqa`?

tests/integration/test_mariadb.py
```diff
@@ -0,0 +1,129 @@
...
+        ),
+        ops_test.model.deploy(app_charm, application_name=APPLICATION_APP_NAME, num_units=2),
+    )
+
+    # Reduce the update_status frequency until the cluster is deployed
+    async with ops_test.fast_forward():
+
+        await ops_test.model.block_until(
+            lambda: len(ops_test.model.applications[DATABASE_APP_NAME].units) == 3
+        )
```
zmraul, line 58: I would consider adding a timeout on every block_until. Timeout defaults to None, so infinite block otherwise? I'll do these changes to my own tests as well :)

src/relations/mysql.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def _on_leader_elected(self, _) -> None:
+        """Handle the leader elected event.
+
+        Retrieves relation data from the peer relation databag and copies
+        the relation data into the new leader unit's databag.
+        """
+        # Skip if the charm is not past the setup phase (config-changed event not executed yet)
+        if not self.charm._is_peer_data_set:
+            return
```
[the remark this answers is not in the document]
paulomach: skip it altogether - this is needed when there's a leadership change after the first one only.

src/relations/mysql.py
paulomach: this kind of delay will only happen on initial run, if the apps are related while still on first deploying (e.g. when packed in a bundle). A single update-status interval will be enough.

tests/integration/test_mariadb.py
paulomach: that's strange. retesting now is passing. maybe some transient issue in flake, idk.. removed it now

tests/integration/test_mariadb.py
paulomach: CRUD, admin and relation data are done by the tester charm.
The scale will be done on another PR

.github/workflows/release.yaml
```diff
@@ -69,10 +69,10 @@ jobs:
         with:
           fetch-depth: 0
       - name: Select charmhub channel
-        uses: canonical/charming-actions/channel@1.0.3
```
MiaAltieri, line 72: Question, this file seems to run `tox -e integration-ha`, but shouldn't it run all integration tests before releasing to charmhub, ie: `integration-test-lxd-mysql`, `integration-test-lxd-database`,  `integration-test-lxd-shared-db`, and `integration-test-lxd-db-router`

tests/integration/helpers.py
```diff
@@ -240,3 +240,24 @@ def is_relation_broken(ops_test: OpsTest, endpoint_one: str, endpoint_two: str)
...
+        "host": credentials["host"],
+        "raise_on_warnings": False,
+    }
+
+    try:
+        with MysqlConnector(config) as cursor:
+            cursor.execute("SELECT 1")
+            return cursor.fetchone()[0] == 1
+    except Exception:
+        return False
```
marcoppenheimer, line 274: Does this ever fail for unknown reasons, but a retry would resolve it?
If so, might be worth adding a retry, otherwise ignore.

marcoppenheimer (approved):
LGTM

.github/workflows/release.yaml
paulomach: yep! forgot to replicate after creating new tests. Good catch

tests/integration/helpers.py
paulomach: it's possible for sure. I'll fix this

tests/integration/helpers.py
```diff
@@ -231,3 +250,41 @@ def is_relation_broken(ops_test: OpsTest, endpoint_one: str, endpoint_two: str)
...
+def instance_ip(model: str, instance: str) -> str:
+    """Translate juju instance name to IP.
+
+    Args:
+        model: The name of the model
+        instance: The name of the instance
+    Returns:
+        The (str) IP address of the instance
+    """
+    output = subprocess.check_output(f"juju machines --model {model}".split())
```
shayancanonical, line 286: minor nit: it may be possible to avoid using `subprocess.check_output` (and instead do `ops_test.juju("machines", "--model", f"{model}"`)

tests/integration/test_relation_mysql_legacy.py
```diff
@@ -0,0 +1,155 @@
...
+ENDPOINT = "mysql"
+
+TEST_USER = "testuser"
+TEST_DATABASE = "testdb"
+
+
+@pytest.mark.order(1)
+@pytest.mark.abort_on_fail
+@pytest.mark.skip_if_deployed
+@pytest.mark.mariadb_tests
```
shayancanonical, line 40: minor nit: `mariadb_tests` -> `mysql_interface_tests`?

shayancanonical (approved):
LGTM, just a couple minor changes suggested

tests/integration/helpers.py
paulomach: That's was my first approach, but it was failing due to a bug in python libjuju, so I've just change to this

MiaAltieri (approved):

MiaAltieri (approved):

PR #39, opened 2022-08-30
Adding get_secret and set_secret placeholders for secret management.

averma-canonical:
# Issue
Make charm code amenable for plugging in juju secret management apis later.

# Solution
Introduce get and set secret apis for all secret access within charm.

# Testing
Unit tests added for get_secret and set_secret


src/charm.py
```diff
@@ -38,6 +39,10 @@
 
 logger = logging.getLogger(__name__)
 
+ROOT_PASSWORD_KEY = "root-password"
+SERVER_CONFIG_PASSWORD_KEY = "server-config-password"
+CLUSTER_ADMIN_PASSWORD_KEY = "cluster-admin-password"
```
shayancanonical, line 44: would these be better placed in `constants.py`?

shayancanonical (approved):
Looks great!

src/charm.py
averma-canonical: Thanks Shayan! Done.

delgod (approved):

Issue #32, opened 2022-08-25
database-relation-broken hook fails when mysql primary unit is removed after relating with application charm


averma-canonical:
### Bug Description

In a charmed mysql cluster (after relating to the test application charm), if primary unit is removed, database-relation-broken hook gets triggered and fails with the following exception.

unit-mysql-1: 11:52:13 ERROR unit.mysql/1.juju-log database:2: Uncaught exception while in charm code:
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-mysql-1/charm/src/mysqlsh_helpers.py", line 172, in _run_mysqlsh_script
    return subprocess.check_output(command, stderr=subprocess.PIPE).decode("utf-8")
  File "/usr/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['/snap/bin/mysqlsh', '--no-wizard', '--python', '-f', '/root/snap/mysql-shell/common/tmpirx2nv08']' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-mysql-1/charm/lib/charms/mysql/v0/mysql.py", line 858, in get_cluster_primary_address
    output = self._run_mysqlsh_script("\n".join(get_cluster_primary_commands))
  File "/var/lib/juju/agents/unit-mysql-1/charm/src/mysqlsh_helpers.py", line 174, in _run_mysqlsh_script
    raise MySQLClientError(e.stderr)
charms.mysql.v0.mysql.MySQLClientError: b'Cannot set LC_ALL to locale en_US.UTF-8: No such file or directory\nTraceback (most recent call last):\n  File "<string>", line 2, in <module>\nRuntimeError: Dba.get_cluster: This function is not available through a session to a standalone instance (metadata exists, instance does not belong to that metadata, and GR is not active)\n\n'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./src/charm.py", line 301, in <module>
    main(MySQLOperatorCharm)
  File "/var/lib/juju/agents/unit-mysql-1/charm/venv/ops/main.py", line 431, in main
    _emit_charm_event(charm, dispatcher.event_name)
[...]

PR #40, opened 2022-08-31
Bugfix/relation broken

paulomach:
# Issue

Refer to issue #32 

# Solution

Check peer access to diagnose if the unit is being torn down or if the relation is really broken.

# Context

Juju generates a `relation-broken` event for a departing unit, even if the relation is not being broken.
This was reported in [this bug](https://bugs.launchpad.net/juju/+bug/1979811).

# Testing

Manual validation using reported scenario. Integration test is going to be added in shakedown tests PR.


shayancanonical (approved):

averma-canonical (approved):
Thanks Paulo for figuring out the juju behavior and addressing this. Looks good to me!

PR #41, opened 2022-09-01
Password rotate action implementation.

averma-canonical:
# Issue
Enhancement to add action to rotate admin passwords.

# Solution
Implementation of actions get_password, set_password

# Context
Need an action to manage credentials for various admin passwords.

# Testing
Integration test added for explicit and silent password rotation scenarios.

### DIFF
--- a/actions.yaml
+++ b/actions.yaml
@@ -11,7 +11,7 @@ get-password:
     username:
       type: string
       description: The username, the default value 'operator'.
-        Possible values - operator, backup, replication.
+        Possible values - root, serverconfig, clusteradmin.
 
 set-password:
   description: Change the system user's password, which is used by charm.
@@ -20,7 +20,7 @@ set-password:
     username:
       type: string
       description: The username, the default value 'operator'.
-        Possible values - operator, backup, replication.
+        Possible values - root, serverconfig, clusteradmin.
     password:
       type: string
       description: The password will be auto-generated if this option is not specified.
--- a/actions.yaml
+++ b/actions.yaml
@@ -12,7 +12,6 @@ get-password:
       type: string
       description: The username, the default value 'operator'.
         Possible values - operator, backup, replication.
-  required: [username]
 
 set-password:
   description: Change the system user's password, which is used by charm.
@@ -25,4 +24,3 @@ set-password:
     password:
       type: string
       description: The password will be auto-generated if this option is not specified.
-  required: [username]
--- a/actions.yaml
+++ b/actions.yaml
@@ -12,3 +12,27 @@ get-root-credentials:
 
 get-cluster-status:
   description: Get cluster status information
+
+get-password:
+  description: Fetch the system user's password, which is used by charm.
+    It is for internal charm users and SHOULD NOT be used by applications.
+  params:
+    username:
+      type: string
+      description: The username, the default value 'operator'.
+        Possible values - operator, backup, replication.
+
+set-password:
+  description: Change the system user's password, which is used by charm.
+    It is for internal charm users and SHOULD NOT be used by applications.
+  params:
+    username:
+      type: string
+      description: The username, the default value 'operator'.
+        Possible values - operator, backup, replication.
+    password:
+      type: string
+      description: The password will be auto-generated if this option is not specified.
+
+configure-backup-storage:
+  description: Configure the backup storage
--- a/actions.yaml
+++ b/actions.yaml
@@ -1,15 +1,6 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
 
-get-cluster-admin-credentials:
-  description: Get the credentials for the cluster admin user
-
-get-server-config-credentials:
-  description: Get the credentials for the server config user
-
-get-root-credentials:
-  description: Get the credentials for the root user
-
 get-cluster-status:
   description: Get cluster status information
 
@@ -21,6 +12,7 @@ get-password:
       type: string
       description: The username, the default value 'operator'.
         Possible values - operator, backup, replication.
+  required: [username]
 
 set-password:
   description: Change the system user's password, which is used by charm.
@@ -33,6 +25,4 @@ set-password:
     password:
       type: string
       description: The password will be auto-generated if this option is not specified.
-
-configure-backup-storage:
-  description: Configure the backup storage
+  required: [username]
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -979,6 +979,30 @@ class MySQLBase(ABC):
             logger.warning(f"Failed to grant privileges to user {username}@{hostname}", exc_info=e)
             raise MySQLGrantPrivilegesToUserError(e.message)
 
+    def update_user_password(
+        self, username: str, new_password: str, current_server_config_password: str
+    ) -> None:
+        """Updates user password in MySQL database."""
+        logger.debug(f"Updating password for {username}.")
+
+        connect_instance_address = self.instance_address
+        server_config_username = self.server_config_user
+
+        update_user_password_commands = (
+            f"shell.connect('{server_config_username}:{current_server_config_password}@{connect_instance_address}')",
+            f"session.run_sql(\"ALTER USER '{username}'@'%' IDENTIFIED BY '{new_password}';\")",
+            'session.run_sql("FLUSH PRIVILEGES;")',
+        )
+
+        try:
+            self._run_mysqlsh_script("\n".join(update_user_password_commands))
+        except MySQLClientError as e:
+            logger.exception(
+                f"Failed to update user password for user {username}",
+                exc_info=e,
+            )
+            raise MySQLCheckUserExistenceError(e.message)
+
     @abstractmethod
     def wait_until_mysql_connection(self) -> None:
         """Wait until a connection to MySQL has been obtained.
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -1006,9 +1006,6 @@ class MySQLBase(ABC):
             )
             raise MySQLCheckUserExistenceError(e.message)
 
-        if username == self.server_config_user:
-            self.server_config_password = new_password
-
     @abstractmethod
     def wait_until_mysql_connection(self) -> None:
         """Wait until a connection to MySQL has been obtained.
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -979,17 +979,20 @@ class MySQLBase(ABC):
             logger.warning(f"Failed to grant privileges to user {username}@{hostname}", exc_info=e)
             raise MySQLGrantPrivilegesToUserError(e.message)
 
-    def update_user_password(
-        self, username: str, new_password: str, current_server_config_password: str
-    ) -> None:
-        """Updates user password in MySQL database."""
+    def update_user_password(self, username: str, new_password: str) -> None:
+        """Updates user password in MySQL database.
+
+        Args:
+            username: The username of user to update the password for
+            new_password: The new password to be set for the user mentioned in username arg
+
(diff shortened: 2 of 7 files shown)


paulomach:
@averma-canonical as we talk, we have a few points to tackle:

* split configure backup into another PR. This should be about password rotation only
* remove deprecated actions for credentials retrieval
  * make sure if these are used in integrations tests, to refactor accordingly
* add system users list to `constants.py`
* if possible refactor get/set secrets without `scope`, since most probably `unit` is not used - double check legacy relations
* extend `mysql.py` lib with a update_user_password function
* replicate PR's to mysql-k8s operator

delgod:
> * if possible refactor get/set secrets without `scope`, since most probably `unit` is not used - double check legacy relations

I would keep `scope`. Reason: we are going to implement TLS integration soon. Each unit needs its own TLS certificate but only leader can write into `scope="app"`

src/charm.py
```diff
@@ -224,37 +218,64 @@ def _on_database_storage_detaching(self, _) -> None:
...
-        )
+    def _on_get_password(self, event: ActionEvent) -> None:
+        """Action used to retrieve the system user's password."""
+        if "username" not in event.params:
+            raise RuntimeError("Undefined parameter username.")
+
+        if event.params["username"] not in REQUIRED_USERNAMES:
+            raise RuntimeError("Invalid username.")
+
+        username = event.params["username"]
```
delgod, line 233: I would set `username=root` as the default username to make UX simpler. and the same for setting password
```suggestion
        username = ROOT_USERNAME
        if "username" in event.params:
            username = event.params["username"]

        if username not in REQUIRED_USERNAMES:
            raise RuntimeError("Invalid username.")

```

src/charm.py
```diff
@@ -224,37 +218,64 @@ def _on_database_storage_detaching(self, _) -> None:
...
+            raise RuntimeError("Invalid username.")
 
-    def _on_get_server_config_credentials(self, event: ActionEvent) -> None:
-        """Action used to retrieve the server config credentials."""
         event.set_results(
-            {
-                "server-config-username": SERVER_CONFIG_USERNAME,
-                "server-config-password": self._get_secret("app", SERVER_CONFIG_PASSWORD_KEY),
-            }
+            {"username": event.params["username"], "password": self._get_secret(secret_key)}
```
delgod, line 244:
```suggestion
            {"username": username, "password": self._get_secret(secret_key)}
```

delgod (approved):

src/charm.py
```diff
@@ -224,37 +220,63 @@ def _on_database_storage_detaching(self, _) -> None:
...
+            raise RuntimeError("Invalid username.")
+
+        event.set_results({"username": username, "password": self._get_secret("app", secret_key)})
+
+    def _on_set_password(self, event: ActionEvent) -> None:
+        """Action used to update/rotate the system user's password."""
+        if not self.unit.is_leader():
+            raise RuntimeError("set-password action can only be run on the leader unit.")
+
+        username = ROOT_USERNAME
```
paulomach, line 253: same as L232

src/charm.py
```diff
@@ -224,37 +220,63 @@ def _on_database_storage_detaching(self, _) -> None:
...
+        if username == ROOT_USERNAME:
+            secret_key = ROOT_PASSWORD_KEY
+        elif username == SERVER_CONFIG_USERNAME:
+            secret_key = SERVER_CONFIG_PASSWORD_KEY
+        elif username == CLUSTER_ADMIN_USERNAME:
+            secret_key = CLUSTER_ADMIN_PASSWORD_KEY
+        else:
+            raise RuntimeError("Invalid username.")
+
+        new_password = None
```
paulomach, line 270: you can replace these lines with this:

```suggestion
        new_password = event.params.get("password") or generate_random_password(PASSWORD_LENGTH)
```

src/charm.py
```diff
@@ -224,37 +220,63 @@ def _on_database_storage_detaching(self, _) -> None:
...
+        if "password" not in event.params or event.params["password"] == "":
+            new_password = generate_random_password(PASSWORD_LENGTH)
+        else:
+            new_password = event.params["password"]
+
+        current_server_config_password = self._get_secret("app", SERVER_CONFIG_PASSWORD_KEY)
```
paulomach, line 276: you should not need to store (or pass to the function) the current password, since you are setting on the databag after changing in mysql.

lib/charms/mysql/v0/mysql.py
```diff
@@ -979,6 +979,30 @@ def grant_privileges_to_user(
             logger.warning(f"Failed to grant privileges to user {username}@{hostname}", exc_info=e)
             raise MySQLGrantPrivilegesToUserError(e.message)
 
+    def update_user_password(
+        self, username: str, new_password: str, current_server_config_password: str
```
paulomach, line 983: you should remove the current password setting here (see comment on `charm.py`)

lib/charms/mysql/v0/mysql.py
```diff
@@ -979,6 +979,30 @@ def grant_privileges_to_user(
...
+    ) -> None:
+        """Updates user password in MySQL database."""
+        logger.debug(f"Updating password for {username}.")
+
+        connect_instance_address = self.instance_address
+        server_config_username = self.server_config_user
```
paulomach, line 989: these are redundant, just use `self.*` directly on the f-strings below

paulomach (changes requested):
Abi, some changes, tackle them and we are good to go

lib/charms/mysql/v0/mysql.py
paulomach: But I don't see why you would need to refresh it, i.e.: 

`actions is called -> new pass to mysql using old pass -> new pass to databag -> hook finish`

next runs of anything will run with new password already

averma-canonical: I am talking about the scenario where 
- we rotate the serverconfig password and then after that 
- try to rotate any other password. 
 
In this scenario, the update_user_password() function which relies on the serverconfig password will fail since it's copy of the serverconfig password loaded at init would be out of sync.

actions.yaml
```diff
@@ -1,14 +1,26 @@
...
   description: Get cluster status information
+
+get-password:
+  description: Fetch the system user's password, which is used by charm.
+    It is for internal charm users and SHOULD NOT be used by applications.
+  params:
+    username:
+      type: string
+      description: The username, the default value 'operator'.
+        Possible values - operator, backup, replication.
```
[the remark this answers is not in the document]
averma-canonical: ack

actions.yaml
```diff
@@ -1,14 +1,26 @@
...
+
+set-password:
+  description: Change the system user's password, which is used by charm.
+    It is for internal charm users and SHOULD NOT be used by applications.
+  params:
+    username:
+      type: string
+      description: The username, the default value 'operator'.
+        Possible values - operator, backup, replication.
```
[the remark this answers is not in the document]
averma-canonical: ack

lib/charms/mysql/v0/mysql.py
averma-canonical: ack

src/charm.py
```diff
@@ -224,37 +220,63 @@ def _on_database_storage_detaching(self, _) -> None:
...
-    def _on_get_server_config_credentials(self, event: ActionEvent) -> None:
-        """Action used to retrieve the server config credentials."""
-        event.set_results(
-            {
-                "server-config-username": SERVER_CONFIG_USERNAME,
-                "server-config-password": self._get_secret("app", SERVER_CONFIG_PASSWORD_KEY),
-            }
-        )
+        if "username" in event.params:
+            username = event.params["username"]
```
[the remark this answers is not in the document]
averma-canonical: I think it will throw KeyError: 'username'

src/charm.py
averma-canonical: I think it will throw KeyError: 'username'

src/charm.py
averma-canonical: I think it will throw KeyError: 'password'

lib/charms/mysql/v0/mysql.py
averma-canonical: ack, will update the lib's copy of serverconfig password as well to keep it in sync

lib/charms/mysql/v0/mysql.py
```diff
@@ -979,6 +979,36 @@ def grant_privileges_to_user(
...
+        try:
+            self._run_mysqlsh_script("\n".join(update_user_password_commands))
+        except MySQLClientError as e:
+            logger.exception(
+                f"Failed to update user password for user {username}",
+                exc_info=e,
+            )
+            raise MySQLCheckUserExistenceError(e.message)
+
+        if username == self.server_config_user:
```
shayancanonical, line 1009: why is this needed? this MySQL class is instantiated anew every time a handler is invoked (since the charm class is also instantiated anew every time a handler is invoked)

tests/integration/test_database.py
```diff
@@ -82,6 +90,131 @@ async def test_build_and_deploy(ops_test: OpsTest):
 @pytest.mark.order(2)
 @pytest.mark.abort_on_fail
 @pytest.mark.database_tests
+async def test_password_rotation(ops_test: OpsTest):
+    """Rotate password and confirm changes."""
+    num_units = len(ops_test.model.applications[DATABASE_APP_NAME].units)
+    random_unit = ops_test.model.applications[DATABASE_APP_NAME].units[num_units - 1]
```
shayancanonical, line 96: you could potentially index the units array by `-1` and achieve the same - but is more readable in my opinion

tests/integration/test_database.py
```diff
@@ -82,6 +90,131 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+
+    await rotate_credentials(
+        unit=primary_unit, username=SERVER_CONFIG_USERNAME, password=new_password
+    )
+
+    updated_credentials = await fetch_credentials(random_unit, SERVER_CONFIG_USERNAME)
+    assert updated_credentials["password"] != old_credentials["password"]
+    assert updated_credentials["password"] == new_password
+
+    primary_unit = await get_primary_unit(
```
shayancanonical, line 123: why do we need to log what node is the primary member on line 133? could we remove lines 123-133?

tests/integration/test_database.py
```diff
@@ -82,6 +90,131 @@ async def test_build_and_deploy(ops_test: OpsTest):
...
+    logger.debug(
+        "Test succeeded Primary unit detected before password rotation is %s", primary_unit_address
+    )
+
+    await rotate_credentials(unit=primary_unit, username=SERVER_CONFIG_USERNAME)
+
+    updated_credentials = await fetch_credentials(random_unit, SERVER_CONFIG_USERNAME)
+    assert updated_credentials["password"] != old_credentials["password"]
+
+    primary_unit = await get_primary_unit(
```
shayancanonical, line 165: same thing as above, do we need lines 165-175 in this test?

lib/charms/mysql/v0/mysql.py
averma-canonical: ack, I was not aware. was being defensive.

tests/integration/test_database.py
averma-canonical: - primary unit is needed to issue set password action.
- it is also an actual test for the sanity of rotated password, of course post rotation

src/charm.py
averma-canonical: ack

tests/integration/test_database.py
averma-canonical: Added comments in tests to explain what I am doing, plus changed the post rotation password check query

tests/integration/test_database.py
averma-canonical: ack

shayancanonical (approved):

paulomach (approved):

PR #43, opened 2022-09-07
Simplify flaky db-router test by reducing number of mysql units

shayancanonical:
# Issue
1. We are deploying a HA mysql cluster for the db-router tests
2. We are testing no loss of data with the mysql cluster scale down in the db-router tests (where we have separate tests that test this)
3. The [following error](https://github.com/canonical/mysql-operator/runs/8232907283?check_suite_focus=true) is likely to occur due to a race condition (deploying mysql and mysqlrouter simultaneously, router could be deployed and related to mysql before mysql is deployed)

# Solution
1. Deploy a cluster with 1 mysql unit
2. Do not test the "no loss of data with the cluster scale down"
3. Refactor to avoid the race condition

# Release Notes
Simplify flaky db-router test by reducing number of mysql units

### DIFF
--- a/tests/integration/test_db_router.py
+++ b/tests/integration/test_db_router.py
@@ -129,7 +129,7 @@ async def test_keystone_bundle_db_router(ops_test: OpsTest) -> None:
         "keystone",
         series="focal",
         application_name=KEYSTONE_APP_NAME,
-        num_units=2,
+        num_units=1,
     )
 
     # Deploy mysqlrouter and relate it to keystone
@@ -196,7 +196,7 @@ async def test_keystone_bundle_db_router(ops_test: OpsTest) -> None:
             "keystone",
             series="focal",
             application_name=ANOTHER_KEYSTONE_APP_NAME,
-            num_units=2,
+            num_units=1,
         )
 
         # Deploy mysqlrouter and relate it to keystone
--- a/tests/integration/test_db_router.py
+++ b/tests/integration/test_db_router.py
@@ -137,6 +137,7 @@ async def test_keystone_bundle_db_router(ops_test: OpsTest) -> None:
     # Deploy mysqlrouter and relate it to keystone
     keystone_mysqlrouter_app = await ops_test.model.deploy(
         "mysql-router",
+        channel="8.0/stable",
         application_name=KEYSTONE_MYSQLROUTER_APP_NAME,
     )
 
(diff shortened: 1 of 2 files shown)


paulomach:
Have you saw any improvement here?

tests/integration/test_db_router.py
```diff
@@ -27,69 +26,8 @@
...
-        f"{keystone_mysqlrouter_application_name}:db-router", f"{APP_NAME}:db-router"
-    )
-    await ops_test.model.block_until(
-        lambda: keystone_app.status in ("active", "error")
-        and keystone_mysqlrouter_app.status in ("active", "error"),
-        timeout=SLOW_WAIT_TIMEOUT,
-    )
-    assert keystone_app.status == "active" and keystone_mysqlrouter_app.status == "active"
+SLOW_WAIT_TIMEOUT = 45 * 60
+FAST_WAIT_TIMEOUT = 30 * 60
```
averma-canonical, line 30: FAST_WAIT_TIMEOUT not used anymore I believe

tests/integration/test_db_router.py
```diff
@@ -239,9 +196,41 @@ async def test_keystone_bundle_db_router(ops_test: OpsTest) -> None:
...
+            "keystone",
+            series="focal",
+            application_name=ANOTHER_KEYSTONE_APP_NAME,
+            num_units=2,
+        )
+
+        # Deploy mysqlrouter and relate it to keystone
+        another_keystone_mysqlrouter_app = await ops_test.model.deploy(
+            "mysql-router",
+            channel="8.0/stable",  # pin to channel as it contains a fix to https://bugs.launchpad.net/charm-mysql-router/+bug/1927981
```
averma-canonical, line 211: I believe this is temporary. Let's raise a placeholder issue to remove this once the fix propagates to latest/default channel.

averma-canonical (approved):
Looks good.

tests/integration/test_db_router.py
shayancanonical: good idea! created a placeholder issue [here](https://github.com/canonical/mysql-operator/issues/45)

tests/integration/test_db_router.py
```diff
@@ -171,60 +109,79 @@ async def check_keystone_users_existence(
...
+            mysql_app.status == "active"
+            and keystone_app.status == "waiting"
+            and keystone_mysqlrouter_app.status == "blocked"
+        )
+
+        # Relate mysqlrouter to mysql
+        await ops_test.model.relate(
+            f"{KEYSTONE_MYSQLROUTER_APP_NAME}:db-router", f"{APP_NAME}:db-router"
+        )
+        await ops_test.model.block_until(
```
WRFitch, line 174: you could use wait_for_idle(), with `status="error"` and `block_on_error=False`. Up to you though

WRFitch (approved):

tests/integration/test_db_router.py
shayancanonical: we did use `wait_for_idle` before, but it apparently takes longer to stabilize after scaling up. so we switched to `block_until` because the only thing we care about is whether the application went to `active` status

PR #48, opened 2022-09-30
Feature TLS implementation

paulomach:
# Issue

Implementation of TLS for mysql-vm, from no support to TLS operator charm integration.
Reference to [DPE-625](https://warthogs.atlassian.net/browse/DPE-625), [DPE-626](https://warthogs.atlassian.net/browse/DPE-626), [DPE-627](https://warthogs.atlassian.net/browse/DPE-627)

# Solution

Use a TLS provider module to implement the tls relation, it's events and workload handling to deal with add/remove tls support into mysql.
Notes:
* using rolling_ops support for cluster restart after changes.
* need for internal/external keys to-be-confirmed
* validation of extra config when using XCOM vs. MySQL comm-modes
  * **NOTE** This feature is only enabled by `mysql-shell >= 8.0.28` - There's a [open MP](https://code.launchpad.net/~paulomachado/snap-mysql-shell/+git/snap-mysql-shell/+merge/431297) to update the mysql-shell snap, which must be merged before this PR - All integration tests will fail while this is not merged.

# Testing

Manual validation only.
Unit and Integration tests to be done on follow-up PRs

# Release Notes

*quick math: Out of 1972 lines added, 1599 are imported libs, and only 373 are stuff that maters*

**Important changes**

* `actions.yaml` - new actions
* `src/charm.py` - rollings_ops manager and callback function, made some methods public
* `src/constants.py` - TLS related constants
* `src/mysqlsh_helpers.py` - new platform (vm) specific write-to-file method
* `src/relations/mysql_tls.py` - TLS provider implementation
* `templates/tls.cnf` - mysql-server template config for TLS enablement

Remaining are library imports, boilerplate config, or some reflection of making class method public (i.e. `s/_//`)


src/charm.py
```diff
@@ -249,12 +256,12 @@ def _on_database_storage_detaching(self, _) -> None:
...
 
         # Inform other hooks of current status
-        self._peers.data[self.unit]["unit-status"] = "removing"
+        self.peers.data[self.unit]["unit-status"] = "removing"
 
     def _on_update_status(self, _) -> None:
         """Handle update status.
 
-        Takes care of workload health checks.
+        Takes care of workload health checks and.
```
WRFitch, line 264: Typo here

src/charm.py
```diff
@@ -444,6 +451,12 @@ def _set_secret(self, scope: str, key: str, value: Optional[str]) -> None:
         else:
             raise RuntimeError("Unknown secret scope.")
 
+    @property
+    def active_status_message(self):
```
WRFitch, line 455: nice

WRFitch (approved):

src/relations/mysql_tls.py
```diff
@@ -0,0 +1,237 @@
...
+    def _on_tls_relation_broken(self, _) -> None:
+        """Disable TLS when TLS relation broken."""
+        self.charm.set_secret(SCOPE, "ca", None)
+        self.charm.set_secret(SCOPE, "cert", None)
+        self.charm.set_secret(SCOPE, "chain", None)
+        self.remove_tls_config_file()
+        # set member-state to avoid unwanted health-check actions
+        self.charm.unit_peer_data["member-state"] = "waiting"
+        # trigger rolling restart
+        self.charm.on[self.charm.restart_manager.name].acquire_lock.emit()
```
shayancanonical, line 126: to confirm my understanding: does setting `communicationStack = MYSQL` just enable the ability to use TLS for XProtocol, but is not activated until the TLS CA, cert and chain are available? do we need to (or can we do) something to move cluster `communicationStack = XCOM`? (Does `communicationStack = MYSQL` fail if TLS files are not present?

src/relations/mysql_tls.py
```diff
@@ -0,0 +1,237 @@
...
+            )
+
+    @staticmethod
+    def create_tls_config_file() -> None:
+        """Render TLS template directly to file.
+
+        Render and write TLS enabling config file from template.
+        """
+        with open("templates/tls.cnf", "r") as template_file:
+            template = Template(template_file.read())
```
shayancanonical, line 216: should we use Jinja instead? (that is what we use for templates in other projects and elsewhere)

src/relations/mysql_tls.py
```diff
@@ -0,0 +1,237 @@
...
+        self.framework.observe(self.certs.on.certificate_expiring, self._on_certificate_expiring)
+
+    # =======================
+    #  Event Handlers
+    # =======================
+    def _on_set_tls_private_key(self, event: ActionEvent) -> None:
+        """Action for setting a TLS private key."""
+        self._request_certificate(event.params.get("internal-key", None))
+
+    def _on_tls_relation_joined(self, _) -> None:
```
shayancanonical, line 68: should this be `_on_tls_relation_created`? based on my understanding `tls_relation_joined` events will be received by MySQL when new units are added to the TLS operator application

src/relations/mysql_tls.py
```diff
@@ -0,0 +1,237 @@
...
+
+        self.framework.observe(self.certs.on.certificate_available, self._on_certificate_available)
+        self.framework.observe(self.certs.on.certificate_expiring, self._on_certificate_expiring)
+
+    # =======================
+    #  Event Handlers
+    # =======================
+    def _on_set_tls_private_key(self, event: ActionEvent) -> None:
+        """Action for setting a TLS private key."""
+        self._request_certificate(event.params.get("internal-key", None))
```
shayancanonical, line 66: is it worth anything to set the `internal-key` into a secret?

src/charm.py
```diff
@@ -487,13 +500,21 @@ def _workload_reset(self) -> StatusBase:
...
 
-        return ActiveStatus()
+        return ActiveStatus(self.active_status_message)
 
     def _get_primary_address_from_peers(self) -> str:
         """Retrieve primary address based on peer data."""
-        for unit in self._peers.units:
-            if self._peers.data[unit]["member-role"] == "primary":
-                return self._peers.data[unit]["instance-hostname"]
+        for unit in self.peers.units:
```
shayancanonical, line 507: i wonder why we have this private helper method: should we be using `_mysql.get_cluster_primary_address()`?

src/relations/mysql_tls.py
zmraul: I think this is the same as we are doing for the rest of the charms. Each unit will have its own pair of private-public keys, so each unit will need to create a a certificate signing request. Every unit will have it's own certificates, but they all come from the same CA (the one that tls-certificates-operator provides)

src/relations/mysql_tls.py
paulomach: I prefer `Template` over jinja because is lightweight and a builtin, more appropriate for simple tag replacement. Not passionate about it tho. Tbh i forgot to check what we were using, but I propose to change elsewhere or here to use a single solution.

src/relations/mysql_tls.py
paulomach: That's exactly it.

zmraul: @shayancanonical Reading your question again, I didn't understand it right. To expand on what I said, I think we are also assuming that `tls-certificates-operator` will only have one unit.

src/relations/mysql_tls.py
paulomach: No. Using mysql as commstack enables encryption (and [other features](https://dev.mysql.com/doc/refman/8.0/en/group-replication-connection-security.html)) by default for the communication between group replication nodes.
It's independent of encryption with clients - so that's why we are enabling it no matter what.
Also, make the implementation simpler, because remove the necessity of having both internal and external keys.

src/relations/mysql_tls.py
paulomach: Jump in too quickly also - As @zmraul we are indeed assuming the TLS charm to be always a single-unit deployment.

src/relations/mysql_tls.py
paulomach: on this charm, there's no jinja dependency - so I would prefer to keep as is

src/relations/mysql_tls.py
paulomach: Yes, and it's done on the `self._request_certificate`

shayancanonical (approved):
I would recommend not merging this in `main` until integration tests exist (merge the integration tests branch into this branch first)

paulomach:
> I would recommend not merging this in `main` until integration tests exist (merge the integration tests branch into this branch first)

Yep, is set to merge to the `tls_support` branch first

PR #61, opened 2022-10-28
Add base of jammy

shayancanonical:
# Issue
1. We have no way of running our suite of integration tests against multiple releases of ubuntu
2. We are not currently supporting `jammy`
3. `ops_test.build_charm` does not support building multiple charm bases

# Solution
1. Use Github CI matrix to run each integration test against a list of `series`. Additonally, use `pytest_generate_test` to add the `series` passed from `tox` to into each test.
2. Add support for `jammy`
3. Manually run `charmcraft pack` in a subprocess to pack charms and then pick the correct archive file to deploy based on the series

# Release Notes
Add base for jammy

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,14 @@
+#!/usr/bin/env python3
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+
+def pytest_addoption(parser):
+    parser.addoption("--series", action="store", type=str, default="jammy")
+
+
+def pytest_generate_tests(metafunc):
+    series = metafunc.config.option.series
+    # Only set "series" if it exists as a fixture for a test
+    if "series" in metafunc.fixturenames:
+        metafunc.parametrize("series", [series])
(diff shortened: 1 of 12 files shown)


paulomach (approved):
lgtm - consider pulling main, since there was some test fixes

conftest.py
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python3
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+
+def pytest_addoption(parser):
+    parser.addoption("--series", action="store", type=str, default="jammy")
+
+
+def pytest_generate_tests(metafunc):
```
WRFitch, line 10: This is neat, I like this

WRFitch (approved):

marceloneppel (approved):
LGTM! Very good idea on how to generate tests for both series!

shayancanonical:
Thanks for the reviews!

PR #65, opened 2022-11-09
Add assumes=juju to avoid starting VM operator on K8s

taurus-forever:
It should avoid human mistakes and negative feedback. Also fix a typo in the Charm summary.

# Issue
It will prevent starting VM operator in K8s env.

# Solution
Using Juju function 'assume'.

# Release Notes
Add protection to avoid starting VM operator on K8s.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -6,7 +6,7 @@ display-name: MySQL
 description: |
   MySQL charm for machines
 summary: |
-  MySQL is  a widely used, open-source relational database management system
+  MySQL is a widely used, open-source relational database management system
   (RDBMS). MySQL InnoDB cluster provides a complete high availability solution
   for MySQL via Group Replication.
 source: https://github.com/canonical/mysql-operator
@@ -39,3 +39,5 @@ storage:
     type: filesystem
     description: Persistent storage for MySQL data
     location: /var/lib/mysql
+assumes:
+  - juju


metadata.yaml
```diff
@@ -39,3 +39,5 @@ storage:
     type: filesystem
     description: Persistent storage for MySQL data
     location: /var/lib/mysql
+assumes:
+  - juju
```
shayancanonical, line 43: im not entirely sure that `assumes=juju` will enforce that only vms are allowed to be deployed. [this post](https://discourse.charmhub.io/t/supported-features/5451) hints at may be using `assumes=juju` to enforce the juju agent version

taurus-forever: It does the job for opposite case:
https://github.com/canonical/mysql-k8s-operator/blob/main/metadata.yaml#L50

```
user@juju1:~$ juju deploy mysql-k8s --channel=edge
Located charm "mysql-k8s" in charm-hub, revision 21
Deploying "mysql-k8s" from charm-hub charm "mysql-k8s", revision 21 in channel edge on focal
ERROR Charm feature requirements cannot be met:
  - charm requires feature "k8s-api" but model does not support it

Feature descriptions:
  - "k8s-api": the Kubernetes API lets charms query and manipulate the state of API objects in a Kubernetes cluster

For additional information please see: https://juju.is/docs/olm/supported-features
user@juju1:~$
```

Should work here as well. :-)

shayancanonical (approved):

PR #69, opened 2022-11-29
Fix typo in pytest mark

carlcsaposs-canonical:


shayancanonical:
@carlcsaposs-canonical you're right, the `dev` tox target is used by paulo to single out an integration test that he'd like to run locally. however, it looks like he just forgot to revert the pytest mark so it is run by the CI suite

shayancanonical (approved):

paulomach (approved):
Exactly how @shayancanonical explained.. sorry and good catch

PR #74, opened 2022-12-12
Fix/blocked after host restart

paulomach:
# Issue

On host system reboot (lxd host, juju controller, vm) a deployed cluster, single or multi units, go into blocked state, as described by in [dpe-907](https://warthogs.atlassian.net/browse/DPE-907).

Also, related to this issue, is that more often than not the attached storage is not being reattached on reboot, as described in [dpe-1007](https://warthogs.atlassian.net/browse/DPE-1007) .

Also, during testing, the issue decribed in [dpe-914](https://warthogs.atlassian.net/browse/DPE-914)  was detected and (TBC) and fixed.

# Solution

Refinements on the initialization handling for better framing the unit state and follow up treatments. Added workaround for flaky lxd storage mount.

# Context

On reboot of the host (or the juju controller for that matter), the charm will receive `on_start` event, which must test if the unit is already initialized, avoiding re-initialization of the workload.

Additionally this change deals with LXD occasionally failing to mount attached storage, i.e. the storage path, as defined in `metadata.yaml`, is not always a mountpoint. This is tested and when the fail is detected, a unit reboot is triggered.

# Testing

Done replicating failures scenarios as described in jira tasks. Integration tests are possible and will be added in future PR.

# Review focus

* `lib/charms/mysql/v0/mysql.py` - add flag for reboot from outage when other nodes are unreachable (or single node case)
* `src/charm.py` - test for storage attached, skip workload initialization when already done
* `src/mysql_vm_helpers.py` - couple of new machine helpers
* `src/relations/db_router.py` and `src/relations/shared_db.py` - avoid running handler when no relations in place (and db calls which can raise errors)
* `tests/unit/test_charm.py` - refactor unit tests so we avoid reboot of the machine running the test


pyproject.toml
```diff
@@ -36,7 +36,7 @@ known_third_party = "mysql.connector"
 [tool.flake8]
 max-line-length = 99
 max-doc-length = 99
-max-complexity = 12
+max-complexity = 13
```
shayancanonical, line 39: im not sure i am completely onboard with this change. we should try using sensible defaults here

src/charm.py
```diff
@@ -103,10 +100,18 @@ def __init__(self, *args):
     #  Charm Lifecycle Hooks
     # =======================
 
-    def _on_install(self, _) -> None:
+    def _on_install(self, event) -> None:
```
shayancanonical, line 103: minor nit: add type hint for event

src/charm.py
```diff
@@ -300,20 +340,19 @@ def _on_update_status(self, _) -> None:
                 self.peers.data[unit].get("member-state", "unknown") for unit in self.peers.units
             }
 
-            if all_states == {"offline"} and self.unit.is_leader():
-                # All instance are off, reboot cluster from outage from the leader unit
-
+            if all_states in [{"offline"}, set()] and self.unit.is_leader():
```
shayancanonical, line 343: @carlcsaposs-canonical  discovered this in the k8s operator, and initialized `all_states = set(['offline'])` as we've already established that the current unit is offline. i like that better as it is more readable

src/charm.py
```diff
@@ -300,20 +340,19 @@ def _on_update_status(self, _) -> None:
...
+                # All instance are off or its a single unit cluster
+                # reboot cluster from outage from the leader unit
                 logger.debug("Attempting reboot from complete outage.")
                 try:
-                    self._mysql.reboot_from_complete_outage()
+                    # reboot from outage forcing it when it a single unit
+                    self._mysql.reboot_from_complete_outage(force=all_states == set())
```
shayancanonical, line 349: it may be easier to understand if `force=len(self.peers.units == 0)`

pyproject.toml
WRFitch: The default on the charm template is 10, so this has already crept up a bit from the norm

src/charm.py
paulomach: this is going out, since we should not never `force`

taurus-forever (approved):
Locally tested and it looks way more stable as previous version.
Good job!

marceloneppel (approved):
Great approach (restarting the unit) for when the unit doesn't have the storage mounted, as we have those issues with lxd. I should copy the idea for PostgreSQL.

taurus-forever (approved):
Nice catch! :-)

PR #77, opened 2022-12-27
Run router bootstrap operations against cluster primary rather than leader unit

shayancanonical:
# Issue
[jira ticket](https://warthogs.atlassian.net/browse/DPE-917)
When running operations against the database while bootstrapping mysqlrouter, the operations are run against the juju leader unit. However, these operations contain writes and the leader unit is not guaranteed to be the cluster primary.

# Solution
Run the operations against the cluster primary unit.

# Release Notes
Run router bootstrap operations against cluster primary rather than leader unit

# Context
The below screenshot demonstrates that the fix in this PR resolves the error.
bootstrap_bugfix


taurus-forever (approved):
Nice!

PR #78, opened 2023-01-03
Cache `charmcraft pack` container in integration tests

### LINKED ISSUE #75: Cache `charmcraft pack` container in integration tests
Cache LXC container used by `charmcraft pack` in GitHub Actions integration tests (to speed up tests)

carlcsaposs-canonical:
# Issue
CI takes a long time to run

# Solution
Cache LXC container used by `charmcraft pack` in GitHub Actions integration tests (to speed up tests)

Closes #75 

# Context
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows

# Testing
Tested on https://github.com/carlcsaposs-canonical/mysql-operator

# Release Notes
Cache LXC container used by `charmcraft pack` in GitHub Actions integration tests (to speed up tests)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,0 +1,121 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+on:
+  workflow_call:
+    inputs:
+      build-matrix:
+        description: JSON string of charms to build
+        required: true
+        type: string
+        # Example:
+        # [
+        #   {
+        #     "job_display_name": "Build mysql charm | focal",
+        #     "bases_index": 0, # corresponds to charmcraft.yaml file
+        #     "path": "."
+        #   },
+        #   {
+        #     "job_display_name": "Build mysql charm | jammy",
+        #     "bases_index": 1,
+        #     "path": "."
+        #   },
+        #   {
+        #     "job_display_name": "Build application charm | jammy",
+        #     "bases_index": 0,
+        #     "path": "tests/integration/relations/application-charm"
+        #   }
+        # ]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        charm: ${{ fromJSON(inputs.build-matrix) }}
+    name: ${{ matrix.charm.job_display_name }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup environment
+        # TODO: Replace with custom image on self-hosted runner
+        run: |
+          # Copied from https://github.com/charmed-kubernetes/actions-operator/blob/96fb0b07eb675f74cf1796be812bc7e67a0d62fc/src/bootstrap/index.ts#L151
+          sudo adduser $USER lxd
+          newgrp lxd
+          sudo lxd waitready
+          sudo lxd init --auto
+          sudo iptables -F FORWARD
+          sudo iptables -P FORWARD ACCEPT
+          sudo chmod a+wr /var/snap/lxd/common/lxd/unix.socket
+          sudo snap install charmcraft --classic
+      - name: Get charmcraft version
+        id: charmcraft-version
+        # Major and minor versions (e.g. "2.1")
+        run: echo "version=$(charmcraft version | cut --delimiter '.' --fields 1,2)" >> $GITHUB_OUTPUT
+      - name: Restore cache of `charmcraft pack` LXC instance
+        id: restore-cache
+        uses: actions/cache/restore@v3
+        with:
+          path: ~/ga-charmcraft-cache/**
+          key: charmcraft-pack-${{ matrix.charm.path }}-${{ matrix.charm.bases_index }}-${{ steps.charmcraft-version.outputs.version }}-${{ hashFiles(format('{0}/charmcraft.yaml', matrix.charm.path), format('{0}/actions.yaml', matrix.charm.path), format('{0}/requirements.txt', matrix.charm.path)) }}
+      - name: Import cached container
+        if: ${{ steps.restore-cache.outputs.cache-hit }}
+        run: |
+          # Project setup copied from https://github.com/canonical/craft-providers/blob/20d154bb8fa9868a678c5621f124a02e2b9e72ad/craft_providers/lxd/project.py#L26
+          lxc project create charmcraft
+          lxc --project default profile show default | lxc --project charmcraft profile edit default
+          charm_repository_directory_inode=$(stat --format "%i" ${{ matrix.charm.path }})
+          for container_tarball in ~/ga-charmcraft-cache/*
+          do
+            lxc --project charmcraft import $container_tarball
+            container_name_without_inode=$(basename --suffix .tar $container_tarball)
+            # Replace placeholder text "INODE" with inode
+            container_name_with_inode=$(echo $container_name_without_inode | sed "s/INODE/$charm_repository_directory_inode/")
+            lxc --project charmcraft move $container_name_without_inode $container_name_with_inode
+          done
+      - name: Pack charm
+        working-directory: ${{ matrix.charm.path }}
+        run: charmcraft pack --bases-index=${{ matrix.charm.bases_index }}
+      - run: touch .empty
+      - name: Upload packed charm
+        uses: actions/upload-artifact@v3
+        with:
+          name: packed-charms
+          # .empty file required to preserve directory structure
+          # See https://github.com/actions/upload-artifact/issues/344#issuecomment-1379232156
+          path: |
+            ${{ matrix.charm.path }}/*.charm
+            .empty
+          if-no-files-found: error
+      - name: Export `charmcraft pack` container to cache
+        run: |
+          mkdir -p ~/ga-charmcraft-cache
+          charm_repository_directory_inode=$(stat --format "%i" ${{ matrix.charm.path }})
+          for container_name_with_inode in $(lxc --project charmcraft list --columns n --format csv)
+          do
+            # Replace inode with placeholder text "INODE"
+            container_name_without_inode=$(echo $container_name_with_inode | sed "s/$charm_repository_directory_inode/INODE/")
+            lxc --project charmcraft move $container_name_with_inode $container_name_without_inode
+            # Use GitHub actions/cache compression
+            lxc --project charmcraft export --optimized-storage --compression none $container_name_without_inode ~/ga-charmcraft-cache/$container_name_without_inode.tar
+          done
+      - if: ${{ github.event_name == 'schedule' && steps.restore-cache.outputs.cache-hit }}
+        name: Delete cache on main
+        # GitHub actions cache is limited to 10 GiB per repository
+        # https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#usage-limits-and-eviction-policy
+        # When the 10 GiB limit is exceeded, GitHub deletes the oldest caches.
+        # If the cache on the main branch is deleted by GitHub,
+        # any new pull requests will be unable to restore a cache.
+        # To avoid that situation, delete the cache on main and save
+        # a new cache with the same key once per day.
+        run: |
+          gh api -H "Accept: application/vnd.github+json" --method DELETE "/repos/{owner}/{repo}/actions/caches?key=${{ steps.restore-cache.outputs.cache-primary-key }}"
+        env:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -10,159 +16,270 @@ jobs:
...
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Install dependencies
-        run: python -m pip install tox
+      - name: Install tox
+        # TODO: Consider replacing with custom image on self-hosted runner OR pinning version
+        run: python3 -m pip install tox
       - name: Run tests
-        run: tox -e unit
+        run: tox run -e unit
```
taurus-forever, line 35: It changes nothing for sure, JFMI, why are you adding it? We use "tox -e unit" in all other repos.

.github/workflows/ci.yaml
```diff
@@ -10,159 +16,270 @@ jobs:
...
+        run: |
+          # Copied from https://github.com/charmed-kubernetes/actions-operator/blob/96fb0b07eb675f74cf1796be812bc7e67a0d62fc/src/bootstrap/index.ts#L151
+          sudo adduser $USER lxd
+          newgrp lxd
+          sudo lxd waitready
+          sudo lxd init --auto
+          sudo iptables -F FORWARD
+          sudo iptables -P FORWARD ACCEPT
+          sudo chmod a+wr /var/snap/lxd/common/lxd/unix.socket
+          sudo snap install charmcraft --classic
```
taurus-forever, line 74: Did you extract this for the speed reason only? Or smth conceptually broken inside charmed-kubernetes/actions-operator@main?

taurus-forever (commented):
1) Wow! I like it! (c)
2) Should we consider to extract matrix / caching logic outside of ci.yaml ? It becomes a monster. Sure a separate topic/PR... maybe new GH Action? CI/CD readability is important.
3) some question inside the review.
4) Included Jarred for CI/CD comments/ideas.

taurus-forever:
> > ```
> > 2. Should we consider to extract matrix / caching logic outside of ci.yaml ? It becomes a monster. Sure a separate topic/PR... maybe new GH Action? CI/CD readability is important.
> > ```
> 
> I experimented with using a reusable workflow and decided to not go that route because (1, primarily) it adds a level of nesting to the UI output of GitHub Actions and makes the logs harder to read/understand and (2) it would require constraining the CI with inputs (e.g. assuming & enforcing a `lint`, `unit-test`, and `integration-` tox environments) and I especially didn't want to constrain the CI when we're still figuring out how to stabilize/mark flaky tests, potentially pin dependencies, and potentially run the tests in smaller modules/more concurrently to speed up CI.

I see your arguments. What about to save python code "name: Prune cache from previous runs" into some script and call it from ci.yaml. It should change nothing on GH UI. Also will allow us to run all python static tests on the script and make ci.yaml better readable.

shayancanonical:
@carlcsaposs-canonical i like  the prune cache logic in a separate file - it makes it more readable and maintainable. if alex agrees, would you mind pulling those changes into this branch (so that we can review them together with the other changes in this PR)?

.github/workflows/ci.yaml
```diff
@@ -1,6 +1,12 @@
+# Copyright 2022 Canonical Ltd.
```
taurus-forever, line 1: 2023? :-D

jardon (commented):
This is a beast of a PR.  I like whats going on here, but I'd like to think theres some simplicity that can be had here.  I have a couple of questions:

1) Is there any way we can utilize something like yq for creating the matrix instead of having to maintain python code for that?

2) Could we split up the workflows that *dont* need the matrix?  

As is, reusability and maintainability are going to suffer.

shayancanonical:
I agree with @jardon. Some feedback:
1. After tracing the code, I feel like I understand the high level picture, but it is hard to verify the small details of why something is done in `ci.yaml` or the prune_cache script. If we can document the flow and some decisions made, i think it would make this more maintainable
2. I wonder if there's a cleaner, simpler way to abstract the caching of lxc containers (since this will need to be adopted across most of our charm repos)

Basically, I'm thinking of how 1) new members to the feature can easily understand what's going on and 2) how to easily port this to other repos

.github/workflows/ci.yaml
```diff
@@ -1,6 +1,12 @@
...
+# See LICENSE file for licensing details.
 name: Tests
+
 on:
   pull_request:
+  push:
+    branches:
+      - main
```
shayancanonical, line 9: i remember these tests intentionally not being run on push to the main branch. this is because the tests are run [here](https://github.com/canonical/mysql-operator/blob/main/.github/workflows/release.yaml#L23-L26) in `release.yaml`

.github/workflows/ci.yaml
```diff
@@ -10,159 +14,160 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Setup operator environment
-        uses: charmed-kubernetes/actions-operator@main
+      - name: Setup environment
+        # TODO: Replace with custom image on self-hosted runner
+        run: |
+          # Copied from https://github.com/charmed-kubernetes/actions-operator/blob/96fb0b07eb675f74cf1796be812bc7e67a0d62fc/src/bootstrap/index.ts#L151
```
shayancanonical, line 64: suggestion: what if all the bash command sections in this file are broken into a directory with bash files (and these files are invoked in the appropriate step)? 

this allows us to create easily readable and easily portable sections of the workflow. additionally, it provides an opportunity to pepper the bash scripts with comments

.github/workflows/prune-cache.py
```diff
@@ -0,0 +1,130 @@
...
+    logging.info(
+        f"{bytes_available / GIBIBYTE:.1f} GiB available, {bytes_required / GIBIBYTE:.1f} GiB required"
+    )
+    if not cache_deletion_order:
+        break
+    # Delete current ref cache if it exists or delete the oldest cache
+    ref, cache = cache_deletion_order.pop(-1)
+    if ref == "refs/heads/main":
+        continue
+    delete_cache(cache)
```
shayancanonical, line 130: thoughts on adding a markdown file somewhere that details how the caching/pruning works? the goal should be allow any new comer to the feature to easily pinpoint what the code does and make any changes necessary

this file is especially daunting for me to trace in detail

.github/workflows/ci.yaml
taurus-forever: Normally it should represent the current year. No need to change it here (now). I will check official Canonical policy first.

taurus-forever:
The "Goal" is a fast tests. "Caching" is only a solution.

Demo explained this PR a lot. Tnx @carlcsaposs-canonical !
Still, "+1" to `CI/CD readability decreased`. (c) @jardon 

We are caching to:
  1) speedup charm building
  2) reuse cached charms for tests

PR should be way smaller IF we:
 1) avoid "building" when unnecessary (e.g. tests app)
 2) avoid using GH cache to share charms with tests

The repo has 60 commits, 44 touched tests and 2 only changed test app "application_charm":

```
> mysql-operator.git$ git log --oneline | wc -l
60
> mysql-operator.git$ git log --oneline tests/ | wc -l
44
> mysql-operator.git$ git log --oneline tests/integration/high_availability/application_charm/ | wc -l
2
```

IDEA1: What about to store `PR charms` in `branches` on `<track>/<risk>/<branch>`, e.g. `latest/edge/pr_78_mysql_123456` `latest/edge/pr_78_app1_654321`. So we will use [snapstore](https://snapcraft.io/docs/publish-to-a-branch) and [charmhub](https://juju.is/docs/olm/deploy-a-charm-from-charmhub#heading--specify-a-charmed-operator-channel) as a cache to share charms with tests. 

Sure, we still need GH caching, but only for `building` charms.

IDEA2: Publish test app on charmhub too. As separate application OR branch `latest/edge/app1_rev42`.
Test will use it if `latest/edge/pr_78_app1_654321` is missing.

Logic could be:

* build main charmcraft.yaml with cache (this PR) for all series (focal, jammy) as matrix. Push to `<branch>` for Tests.
* build test app charmcraft.yaml only if some REVISION is bumped manually (2 commits our of 60 in this repo).
* teach tests to use `<track>/<risk>/<branch>` with fallback to `<track>/<risk>` if no charm available in `<branch>`.
* publish charm to `<track>/<risk>` after merging (like now) and clean `<branch>`.

The `branch` could be `pr_$id_$name_$commit`:
  pr_ : some prefix
  $id : GH PR id
  $name : app name: mysql , app1 , app2
  $commit : git commit id (avoid rebuilding the same commit on re-trigger, etc)

What I like here:
* all charms are stored in stores only. And tests are using all charms from store.
* not-yet merged charms stored in branches. Future: PR tests in `mysql-bundle` can use not-yet-merged `mysql-operator` from `<branch>` (through some `topic` label).
* separation of building / using logic => should simplify this PR.

What do you think?

.github/workflows/ci.yaml
```diff
@@ -10,159 +14,160 @@ jobs:
...
+    outputs:
+      build-matrix: ${{ steps.create-matrix.outputs.build_matrix }}
 
-  integration-test-lxd-db-router:
+  build:
     strategy:
       matrix:
-        series: [focal, jammy]
-    name: Integration tests for db-router relation (lxd)
+        charm: ${{ fromJSON(needs.get-build-matrix.outputs.build-matrix) }}
```
delgod, line 53: It looks too complex to use generated matrix here. I would have a matrix explicitly configured here without scripts.
Any reason to have it generated?

tests/integration/conftest.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
```
delgod, line 1: just Q: where do we use this file?

shayancanonical:
Thinking of simplifying, I was hoping to discuss the following topics to see if they can help make caching more simple/efficient:

1. from my understanding, `charmcraft pack` reuses LXD containers which need to be cleaned when the following files are modified: metadata.yaml, charmcraft.yaml, requirements.txt and actions.yaml. These files do not change very frequently, so I am wondering if we should compile our cache key using these. when the above files are not changed, the lxc container can be reused to pack any file changes in the charm code.
2. [The following discourse post](https://discourse.charmhub.io/t/charmcraft-bases-provider-support/4713) and the [following lxd documentation](https://linuxcontainers.org/lxd/docs/master/remotes/) indicate that we can set up lxd to use a remote server. I am wondering if using a remote machine with sufficient memory/disk space and a cron to periodically clean lxd projects may make caching simpler

Happy to discuss these offline in more detail (and continuing fleshing out the above topics)

jardon:
Could we at least break apart the caching bits into a separate Github Actions repo that is reusable?  That way we can version control it from a central location and each following PR wont be hundreds of lines of code.

If there's any technical reason this is not possible, lmk.

.github/workflows/build.yaml
```diff
@@ -0,0 +1,121 @@
...
+        id: charmcraft-version
+        # Major and minor versions (e.g. "2.1")
+        run: echo "version=$(charmcraft version | cut --delimiter '.' --fields 1,2)" >> $GITHUB_OUTPUT
+      - name: Restore cache of `charmcraft pack` LXC instance
+        id: restore-cache
+        uses: actions/cache/restore@v3
+        with:
+          path: ~/ga-charmcraft-cache/**
+          key: charmcraft-pack-${{ matrix.charm.path }}-${{ matrix.charm.bases_index }}-${{ steps.charmcraft-version.outputs.version }}-${{ hashFiles(format('{0}/charmcraft.yaml', matrix.charm.path), format('{0}/actions.yaml', matrix.charm.path), format('{0}/requirements.txt', matrix.charm.path)) }}
+      - name: Import cached container
```
shayancanonical, line 61: to confirm: if the CI is scheduled (nightly), we will import the cached LXD container for `main`, run charmcraft pack, delete cached LXD container for `main`, and then save the LXD container from the charmcraft pack step?

if so, i have the following suggestion: we should not import any cached LXD containers if scheduled, run charmcraft pack, then save the LXD container from the charmcraft pack step. This will result in a fresh LXD container to be created and cached

shayancanonical (approved):
LGTM, thanks for iterating!

.github/workflows/ci.yaml
taurus-forever: I think this thread can be resolved. Re-checking all threads to submit this PR.

PR #84, opened 2023-01-12
Run CI daily

carlcsaposs-canonical:


taurus-forever (approved):
I am not sure about the proper time 00:53 or 00:42 :-) but let's check how it flies.

shayancanonical (approved):

PR #87, opened 2023-01-17
Use module paths instead of pytest marks to select tests in tox.ini

carlcsaposs-canonical:
# Issue

Integration tests use pytest marks to select a Python module to test

# Solution

Select the Python module directly in tox.ini

# Context

# Testing

# Release Notes

Use module paths instead of pytest marks to select tests in tox.ini

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -62,7 +62,7 @@ deps =
     -r{toxinidir}/requirements.txt
 commands =
     coverage run --source={[vars]src_path},{[vars]lib_path} \
-        -m pytest --ignore={[vars]tests_path}integration -v --tb native -s {posargs}
+        -m pytest -v --tb native -s {posargs} {[vars]tests_path}/unit
     coverage report
 
 [testenv:integration-ha]


tox.ini
```diff
@@ -96,7 +96,7 @@ deps =
     pytest-operator
     -r{toxinidir}/requirements.txt
 commands =
-    pytest -v --tb native --ignore={[vars]tst_path}unit --log-cli-level=INFO -s {posargs} -m "shared_db_tests"
+    ppytest -v --tb native --log-cli-level=INFO -s {posargs} tests/integration/relations/test_shared_db.py
```
shayancanonical, line 99: typo?

shayancanonical (approved):
LGTM, pending one comment. Also, I would recommend using the PR template to fill out the Issue/Solution/Summary/etc. even for minor changes (it keeps the PRs uniform)

shayancanonical:
Thanks for updating. I dont think that there is anything lacking in the information you provided. Also, it seems a bit too much for self-explanatory PRs, but I think that it's still good practice. Thanks!

taurus-forever:
> I do not mind to merge this, but can you please explain which issue we are fixing there? Using marks is not the best solution, maybe, but currently identical between all Git repos. PR description contains "Integration tests use pytest marks to select a Python module to test" which is not an issue you are solving, but the way your have choosen. Tnx!

OK, I think I got it, Just a decreasing amount of marks to mark stable/flaky tests in the upcoming https://github.com/canonical/mysql-operator/pull/88/commits/b7502eb8e0116e16bcbdb80b441b35b2a6d25cfc

LGTM. Tnx!

PR #106, opened 2023-02-06
Move integration test caching to reusable workflow

carlcsaposs-canonical:
# Issue

Caching workflow cannot easily be shared between repositories

# Solution

Move workflow to https://github.com/canonical/data-platform
Depends on https://github.com/canonical/data-platform-workflows/pull/1

# Context

#78 

# Testing

Check github actions output

# Release Notes

Move integration test caching to reusable workflow

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,8 +36,7 @@ jobs:
 
   build:
     name: Build charms
-    # TODO: switch to main branch
-    uses: canonical/data-platform-workflows/.github/workflows/build_charms_with_cache.yaml@feature/build-charms
+    uses: canonical/data-platform-workflows/.github/workflows/build_charms_with_cache.yaml@main
 
   integration-test:
     strategy:


paulomach (approved):
looking nice

tests/integration/conftest.py
```diff
@@ -2,62 +2,69 @@
...
-                # Application charm version does not match mysql charm version
-                # Use latest available version for application charm
-                version = available_versions[-1]
-            return f"local:./{charm_path/charm_name}_ubuntu-{version}-amd64.charm"
+        packed_charms = json.loads(os.environ["CI_PACKED_CHARMS"])
+
+        async def _build_charm(charm_path, bases_index: int = None) -> Path:
+            for charm in packed_charms:
+                if Path(charm_path) == Path(charm["directory_path"]):
+                    if bases_index is None or bases_index == charm["bases_index"]:
```
shayancanonical, line 52: minor nit: the two conditions can be combined with an `and`

shayancanonical (approved):
This is great! LGTM

It might be nice to create a small document that outlines exactly what one needs to do to adopt the caching in their repository

paulomach (approved):
let's go!

tests/integration/conftest.py
shayancanonical: suggestion: we could put () around the second statement

taurus-forever (approved):
I like it! (c)

Issue #104, opened 2023-02-03
Speed up integration tests


carlcsaposs-canonical:
Integration tests currently take 40+ minutes ([example](https://github.com/canonical/mysql-operator/actions/runs/4069205491))

Caching the `charmcraft pack` LXC container (#78) decreased the time by about 15-20 minutes

This issue is to document efforts to further decrease the time

paulomach:
Closing as non topical. Thanks for all the good work here @carlcsaposs-canonical

PR #109, opened 2023-02-10
Parallelize groups of integration tests in a Python module

carlcsaposs-canonical:
## Issue
Integration tests are slow (#104)

## Solution
Group tests in a Python test file with `pytest.mark.group(#)`
Run each group number in a separate GitHub runner

### DIFF
--- a/lib/charms/data_platform_libs/v0/data_interfaces.py
+++ b/lib/charms/data_platform_libs/v0/data_interfaces.py
@@ -16,7 +16,7 @@
 
 This library contains the Requires and Provides classes for handling the relation
 between an application and multiple managed application supported by the data-team:
-MySQL, Postgresql, MongoDB, Redis,  and Kakfa.
+MySQL, Postgresql, MongoDB, Redis, and Kafka.
 
 ### Database (MySQL, Postgresql, MongoDB, and Redis)
 
@@ -303,7 +303,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 7
+LIBPATCH = 9
 
 PYDEPS = ["ops>=2.0.0"]
 
@@ -661,6 +661,11 @@ class DatabaseRequiresEvent(RelationEvent):
     """Base class for database events."""
 
     @property
+    def database(self) -> Optional[str]:
+        """Returns the database name."""
+        return self.relation.data[self.relation.app].get("database")
+
+    @property
     def endpoints(self) -> Optional[str]:
         """Returns a comma separated list of read/write endpoints."""
         return self.relation.data[self.relation.app].get("endpoints")
@@ -742,6 +747,18 @@ class DatabaseProvides(DataProvides):
         # extra user roles) was added to the relation databag by the application.
         if "database" in diff.added:
             self.on.database_requested.emit(event.relation, app=event.app, unit=event.unit)
+
+    def set_database(self, relation_id: int, database_name: str) -> None:
+        """Set database name.
+
+        This function writes in the application data bag, therefore,
+        only the leader unit can call it.
+
+        Args:
+            relation_id: the identifier for a particular relation.
+            database_name: database name.
+        """
+        self._update_relation_data(relation_id, {"database": database_name})
 
     def set_endpoints(self, relation_id: int, connection_strings: str) -> None:
         """Set database primary connections.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -34,7 +34,7 @@ source venv/bin/activate
 tox run -e format        # update your code according to linting rules
 tox run -e lint          # code style
 tox run -e unit          # unit tests
-tox run -m integration   # integration tests
+tox run -e integration   # integration tests
```
paulomach, line 37: eagle eyes

pyproject.toml
```diff
@@ -12,10 +12,7 @@ show_missing = true
 minversion = "6.0"
 log_cli_level = "INFO"
 asyncio_mode = "auto"
-markers = [
```
paulomach, line 15: RIP `dev`

tests/integration/conftest.py
```diff
@@ -1,45 +1,88 @@
...
+        name = function.module.__name__
+        assert name.split(".")[0] == "integration"
+        # Example: "tests/integration/relations/test_database.py"
+        path_to_test_file = f"tests/{name.replace('.', '/')}.py"
+        # Example: "relations/test_database.py | group 1"
+        job_name = f"{'/'.join(path_to_test_file.split('/')[2:])} | group {group_number}"
+        groups.add(Group(path_to_test_file, group_number, job_name))
+    sorted_groups: list[dict] = [dataclasses.asdict(group) for group in sorted(list(groups))]
+    output = f"groups={json.dumps(sorted_groups)}"
+    print(f"\n\n{output}\n")
```
paulomach, line 55: still need this print?

paulomach (commented):
I'm not sure if the added complexity is worth the benefits, in a way I'm not sure if this kind o solutions reaches the point of diminishing returns.
I'm happy to whatever others are on this matter.

paulomach:
> 

Not suggesting. I like the performance gain of course, it's just my future maintenance anxiety kicking in.
I'd like to hear other opinions. @shayancanonical @taurus-forever

shayancanonical:
I agree with Paulo, a 2x speed up in performance is amazing. However, I am also wary of the maintenance cost. One other point not mentioned is that we now have a few tests that are not grouped - if we decide to move this direction, we should at some point change those tests to use this framework as well. Additionally, I'd like to preserve a way to run tests locally (and so would appreciate some documentation for this)

taurus-forever:
1) The incredible job done! Respect, Carl!
2) The speedup [40 mins](https://github.com/canonical/mysql-operator/actions/runs/4309605453/jobs/7517220468) -> [17](https://github.com/canonical/mysql-operator/actions/runs/4287208932/jobs/7467826110)-[25](https://github.com/canonical/mysql-operator/actions/runs/4287208932/jobs/7467826363) mins worth to continue here! 
3) BTW, why difference is so high 25-17=8 mins! Could be a slow GH runner... or not-optimal charm/test is waiting for the next `update-status`? IMHO jammy and focal tests should take more/less the same time. Isn't it?
4) Unfortunately, the old approach is not scalable. The 40 mins will grow after adding new testcases to `test_replication.py`. We need an ability to add new tests/test-cases without slowing down PRs testing. Yes, 60 GH runners is a current problem, but we will solve HW issue.

What are other options to decrease maintenance anxiety?
What is the negative impact of `split the existing test files into multiple files` with a self-explanatory filenames?

taurus-forever:
> FYI I've noticed that sometimes the setup (charmed-kubernetes/actions-operator) can take 5 minutes or sometimes closer to 15 minutes—but maybe that's an unrelated issue.

I saw it today too, [here](https://github.com/canonical/mysql-router-operator/actions/runs/4314481352/jobs/7527690631):
```
Thu, 02 Mar 2023 13:35:40 GMT  Run charmed-kubernetes/actions-operator@main
Thu, 02 Mar 2023 13:35:40 GMT  Install core snap
Thu, 02 Mar 2023 13:35:40 GMT   /usr/bin/sudo snap install core
Thu, 02 Mar 2023 13:44:34 GMT   core 16-2.58.2 from Canonical** installed
```
No ideas about the reasoning.


Re: tests, let's discuss in a sync today.

taurus-forever (approved):
This repos is kind of experimental.
We definitely need faster tests CI/CD. I would vote to independent files as we do not share a lot, but marks are fine too. LGTM to have progress here.

shayancanonical (approved):

PR #117, opened 2023-02-14
Use output from build workflow for artifact name

carlcsaposs-canonical:
## Issue
Artifact name is hard-coded and needs to match with artifact name used during build

## Solution
Use output from build workflow added in https://github.com/canonical/data-platform-workflows/pull/6


paulomach (approved):
:1st_place_medal:

delgod (approved):

shayancanonical (approved):

PR #135, opened 2023-03-01
Bump juju from 2.9.29 to 2.9.38 (latest)

taurus-forever:
## Issue
We are using old Juju.

## Solution
Let's use the latest Juju 2.9.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -87,6 +87,7 @@ jobs:
         uses: charmed-kubernetes/actions-operator@main
         with:
           provider: lxd
+          bootstrap-options: "--agent-version 2.9.38"
       - name: Download packed charm(s)
         uses: actions/download-artifact@v3
         with:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -87,7 +87,7 @@ jobs:
         uses: charmed-kubernetes/actions-operator@main
         with:
           provider: lxd
-          bootstrap-options: "--agent-version 2.9.29"
+          bootstrap-options: "--agent-version 2.9.38"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -87,7 +87,7 @@ jobs:
         uses: charmed-kubernetes/actions-operator@main
         with:
           provider: lxd
-          bootstrap-options: "--agent-version 2.9.29"
+          bootstrap-options: "--agent-version 2.9.38"
```
shayancanonical, line 90: thoughts on dropping the `bootstrap-options` config? i believe [this](https://github.com/charmed-kubernetes/actions-operator/blob/main/src/bootstrap/index.ts#L144) would bootstrap the latest juju version (3.1.0?)

shayancanonical (approved):

shayancanonical (approved):

.github/workflows/ci.yaml
taurus-forever: I see it otherwise, we should always use the latest Juju patchset of 2.9 release.
Also, we should detect ASAP the issues with newly released Juju
(while juju team is in context of solving regressions of newly released component).
I do not see the silent upgrades possibilities, as juju release is loudly announced.

Yes, there is a chance to restore pinning right after 2.9.39 release.
Let's see it on the practice.

taurus-forever: After checking the status in all our repos and the previous experience... I have changed my mind.
Let's keep juju-agent pinned until we trust in our tests stability to avoid PRs blocking in a day of Juju release.

paulomach (approved):
Agreed on keeping it pinned until we sort out and get confident with all tests.

PR #137, opened 2023-03-03
Fix early legacy relations handling

paulomach:
## Issue

Legacy relations are trying to access database (user checking/creation, etc) as soon as the relation starts.
The problem arises when in bundle deployment, the database is still initializing.
Ref. issue is [DPE-1312](https://warthogs.atlassian.net/browse/DPE-1312)

## Solution

Defer relations {created/changed} handlers execution until database is initialized.

[DPE-1312]: https://warthogs.atlassian.net/browse/DPE-1312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -975,10 +975,9 @@ class MySQLBase(ABC):
         Returns:
             The address of the cluster's primary
         """
-        logger.debug(f"Getting cluster primary member's address from {connect_instance_address}")
-
         if not connect_instance_address:
             connect_instance_address = self.instance_address
+        logger.debug(f"Getting cluster primary member's address from {connect_instance_address}")
 
         get_cluster_primary_commands = (
             f"shell.connect('{self.cluster_admin_user}:{self.cluster_admin_password}@{connect_instance_address}')",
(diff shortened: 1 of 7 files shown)


lib/charms/mysql/v0/mysql.py
```diff
@@ -975,10 +975,9 @@ def get_cluster_primary_address(self, connect_instance_address: str = None) -> s
         Returns:
             The address of the cluster's primary
         """
-        logger.debug(f"Getting cluster primary member's address from {connect_instance_address}")
-
         if not connect_instance_address:
             connect_instance_address = self.instance_address
```
shayancanonical, line 979: do we need to patch the LIBPATCH version?

shayancanonical (approved):

lib/charms/mysql/v0/mysql.py
paulomach: i didn't because this does nothing functional. It will eventually be picked up by some update.
I've got annoyed by the messages:
`Getting cluster primary member's address from None`

taurus-forever (approved):

PR #142, opened 2023-03-11
Fix bundle init issues with TLS setup at runtime

paulomach:
## Issue

TLS enablement requires mysqld restarts which lead to unnecessary primary changes/downtime/race conditions. 

[DPE-1420](https://warthogs.atlassian.net/browse/DPE-1420)
and
[DPE-1311](https://warthogs.atlassian.net/browse/DPE-1311)

## Solution

**Moved TLS to charm lib** (so it can be shared with k8s)
Using `ALTER INSTANCE RELOAD TLS` feature.
Removed all need for rolling_ops/restart (which will eventually be required once there's a need to support plugins).

### Chore

Bump libs

[DPE-1420]: https://warthogs.atlassian.net/browse/DPE-1420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DPE-1311]: https://warthogs.atlassian.net/browse/DPE-1311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/charms/mysql/v0/mysql.py
```diff
@@ -1252,6 +1261,70 @@ def offline_mode_and_hidden_instance_exists(self) -> bool:
...
+        except MySQLClientError:
+            raise MySQLTLSRestoreDefaultConfigError
+
+    def kill_unencrypted_sessions(self) -> None:
+        """Kill non local, non system open unencrypted connections."""
+        kill_connections_command = (
+            f"shell.connect('{self.server_config_user}:{self.server_config_password}@{self.instance_address}')",
+            (
+                'processes = session.run_sql("'
+                "SELECT processlist_id FROM performance_schema.threads WHERE "
```
paulomach, line 1314: [case you're wondered where I've got this from](https://dev.mysql.com/doc/refman/8.0/en/performance-schema-threads-table.html)

lib/charms/mysql/v0/mysql.py
```diff
@@ -1252,6 +1261,70 @@ def offline_mode_and_hidden_instance_exists(self) -> bool:
...
+                "connection_type = 'TCP/IP' AND type = 'FOREGROUND';"
+                '")'
+            ),
+            "process_id_list = [id[0] for id in processes.fetch_all()]",
+            'for process_id in process_id_list:\n  session.run_sql(f"KILL CONNECTION {process_id}")',
+        )
```
shayancanonical, line 1320: would this result in client queries throwing an exception? previously, we restarted the instance (which i would expect would terminate connections when it is idle)

paulomach: Client connections will get dropped, in the same way a connection lost would do (and the same way restarting the daemon would for that matter) - so nothing changes from connected apps pov.

shayancanonical: I was under the impression that when we restart the daemon, it will have a grace period to handle SIGTERM (where it can gracefully close open connections). Is this incorrect? If not, then the above killing of threads is different than the restarting the daemon approach

paulomach: The gracefull stop of an open connection has no difference from killing it, since the doc states:

> [KILL CONNECTION](https://dev.mysql.com/doc/refman/8.0/en/kill.html) is the same as [KILL](https://dev.mysql.com/doc/refman/8.0/en/kill.html) with no modifier: It terminates the connection associated with the given processlist_id, after terminating any statement the connection is executing.

So it take effects only when idle

taurus-forever (commented):
It looks very promising! TLS test speedup 18 -> 12 minutes. Nice!

BTW, you most probably need to update [tutorial](https://github.com/canonical/mysql-operator/blob/main/docs/tutorial.md#add-external-tls-certificate) as you are not producing custom TLS files anymore.

LGTM in general, requested Mykola expertise here to avoid planting bomb we might miss.

P.S. Well done!

tests/integration/test_tls.py
```diff
@@ -202,11 +196,8 @@ async def test_disable_tls(ops_test: OpsTest) -> None:
         f"{app}:certificates", f"{TLS_APP_NAME}:certificates"
     )
 
-    # Wait for hooks start reconfiguring app
-    await ops_test.model.block_until(
-        lambda: ops_test.model.applications[app].status != "active", timeout=4 * 60
-    )
-    await ops_test.model.wait_for_idle(apps=[app], status="active", timeout=15 * 60)
+    # Allow time for reconfigure
+    sleep(30)
```
delgod, line 200: Can we test that certificate really changed?
e.g. following command should be able to fetch CA
```
openssl s_client -starttls mysql -connect thehost:3306
```

paulomach: noice, will include that

lib/charms/mysql/v0/mysql.py
```diff
@@ -1252,6 +1261,70 @@ def offline_mode_and_hidden_instance_exists(self) -> bool:
...
+        try:
+            self._run_mysqlcli_script(
+                enable_commands,
+                user=self.server_config_user,
+                password=self.server_config_password,
+            )
+        except MySQLClientError:
+            raise MySQLTLSSetCustomConfigError("Failed to set custom TLS configuration")
+
+    def tls_restore_default(self) -> None:
```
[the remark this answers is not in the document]
paulomach: d88a204

tests/integration/test_tls.py
paulomach: Included on the test app [PR#1](https://github.com/canonical/mysql-test-app/pull/3/)

paulomach:
> 
> BTW, you most probably need to update [tutorial](https://github.com/canonical/mysql-operator/blob/main/docs/tutorial.md#add-external-tls-certificate) as you are not producing custom TLS files anymore.

Actually the files are still generated - they are still required - the path has changed though since we are using the fat snap

src/mysql_vm_helpers.py
```diff
@@ -331,6 +331,29 @@ def reconfigure_mysqld(self) -> None:
         logger.debug("Installing charmed-mysql snap")
         charmed_mysql.ensure(snap.SnapState.Latest, channel="8.0/edge")
 
+    @staticmethod
+    def write_content_to_file(
```
shayancanonical, line 335: should we create an abstract method in the `mysql.py` library for this function?

docs/tutorial.md
```diff
@@ -538,13 +538,13 @@ Like before, connect to the MySQL in one of described above ways and check the T
...
 ```shell
-juju ssh mysql/leader sudo openssl x509 -noout -text -in /var/lib/mysql/custom-server-cert.pem | grep Issuer
+juju ssh mysql/leader sudo openssl x509 -noout -text -in /var/snap/charmed-mysql/common/mysql/data/custom-server-cert.pem | grep Issuer
 ```
 The output should indicate CA configured during TLS operator deployment:
 ```
 Issuer: C = US, CN = Tutorial CA
 ```
-Congratulations! MySQL is now using TLS cetrificate generated by the external application `tls-certificates-operator`.
+Congratulations! MySQL is now using TLS certificate generated by the external application `tls-certificates-operator`.
```
taurus-forever, line 547: Tnx. Nice catch. Curious, why codespell didn't catch this?

taurus-forever (approved):
Cool!

src/mysql_vm_helpers.py
paulomach: not yet - but will eventually. It will require a bit more thought, as in a dedicated PR

PR #165, opened 2023-03-27
Disable check libs label

carlcsaposs-canonical:
PRs already have a status check for the `Check libraries` job; no need to also add labels

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -57,6 +57,7 @@ jobs:
           credentials: "${{ secrets.CHARMHUB_TOKEN }}" # FIXME: current token will expire in 2023-07-04
           github-token: "${{ secrets.GITHUB_TOKEN }}"
           use-labels: false
+          fail-build: true
 
   build:
     name: Build charms


taurus-forever:
I could be wrong, but right now libraries here are outdated, but test is happy and no red label anymore...
What is your vision?

shayancanonical (approved):

taurus-forever:
> > I could be wrong, but right now libraries here are outdated, but test is happy and no red label anymore... @carlcsaposs-canonical what is your vision?
> 
> Good catch; didn't realize the default was for the check to pass if libraries are out of date. Fixed in [fd9d813](https://github.com/canonical/mysql-operator/commit/fd9d813073fde0d81c07c9394d7e9c53ffc4e601)

I had no chance to reply here earlier and unfortunately this is merged.
Are we really willing to enforce us update libraries in all PRs? CC: @carlcsaposs-canonical 

AFAIK, this will affect releasing process as libraries are being updated constantly.
So merged PR will re-execute tests on release and check-lib will block release as some library is outdated for last 2 seconds...

PR #181, opened 2023-04-06
Pin the charmed-mysql snap version

shayancanonical:
## Issue
- We are curriently installing the `8.0/edge` channel of the `charmed-mysql` snap. However, this can cause the charm to break if a new version of the snap is release that is buggy. Additionally, the current set up can result in different versions of MySQL being installed on different nodes of a cluster (if a node is launched after the snap revision for a channel is updated) if a snap updates the MySQL version

## Solution
- Pin the revision of the `charmed-mysql` snap instead of downloading the from the `8.0/edge` channel (revision 44 which is the latest `8.0/edge revision)


shayancanonical:
> (for the future, not necessarily this PR) thinking about how to keep this up to date—will the revision change infrequently enough that it makes sense to update manually or should we look into developing a renovate config to automatically update this?

A renovate config to auto update the revision may be a good idea

paulomach (approved):
lgtm

taurus-forever (approved):

PR #188, opened 2023-04-13
DPE-1656 Limit tests RAM usage by every unit/container to 1Gb

taurus-forever:
## Issue
Often and strange tests crashes on GH runner while all is OK on localhost.

## Solution
GH Runners have 8GB RAM only. We are launching 1-2 MySQL clusters with three units each and allocating 50% for innodb_buffer_pull_size. OOM is an often trouble in this case.


taurus-forever:
Abandoning as we have to go another direction (unfortunately): https://warthogs.atlassian.net/browse/DPE-1656?focusedCommentId=225625

PR #190, opened 2023-04-14
Update README to be in sync with K8s version + update relations/edpoints

taurus-forever:
## Issue
The VM README file was not in sync with K8s one and had no endpoints description, causing issues like https://github.com/canonical/mysql-k8s-operator/issues/198

## Solution
Update README to match the one in mysql K8s repo.

We should consider for the future:
 * name endpoints and interfaces identically/similar
 * sync amount of interfaces supported by VM and K8s


shayancanonical (approved):

paulomach (approved):
One little detail, otherwise looks great

PR #192, opened 2023-04-14
Pin python dependencies with poetry

### LINKED ISSUE #95: Pin pip dependencies with poetry
As implemented by https://github.com/canonical/mongodb-k8s-operator

Purpose: stabilize charm, test dependency updates in isolation from changes to repository

carlcsaposs-canonical:
## Issue
Closes #95
[DPE-1198](https://warthogs.atlassian.net/browse/DPE-1198)

[DPE-1198]: https://warthogs.atlassian.net/browse/DPE-1198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,7 @@ on:
 jobs:
   lint:
     name: Lint
-    uses: canonical/data-platform-workflows/.github/workflows/lint.yaml@v4.0.0
+    uses: canonical/data-platform-workflows/.github/workflows/lint.yaml@v4.1.0
 
   unit-test:
     name: Unit tests
@@ -53,7 +53,7 @@ jobs:
 
   build:
     name: Build charms
-    uses: canonical/data-platform-workflows/.github/workflows/build_charms_with_cache.yaml@v4.0.0
+    uses: canonical/data-platform-workflows/.github/workflows/build_charms_with_cache.yaml@v4.1.0
     with:
       charmcraft-snap-revision: 1349  # version 2.3.0
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,8 +61,8 @@ tenacity = "^8.2.2"
 boto3 = "^1.28.11"
 pyyaml = "^6.0"
 urllib3 = "^1.26.16"
-pytest-operator-cache = {git = "https://github.com/canonical/data-platform-workflows", tag = "v4.0.0", subdirectory = "python/pytest_plugins/pytest_operator_cache"}
-pytest-operator-groups = {git = "https://github.com/canonical/data-platform-workflows", branch = "pytest-plugin-groups", subdirectory = "python/pytest_plugins/pytest_operator_groups"}
+pytest-operator-cache = {git = "https://github.com/canonical/data-platform-workflows", tag = "v4.1.0", subdirectory = "python/pytest_plugins/pytest_operator_cache"}
+pytest-operator-groups = {git = "https://github.com/canonical/data-platform-workflows", tag = "v4.1.0", subdirectory = "python/pytest_plugins/pytest_operator_groups"}
 
 
 [tool.coverage.run]
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -32,14 +32,14 @@ jobs:
 
   build:
     name: Build charm
-    uses: canonical/data-platform-workflows/.github/workflows/build_charm_without_cache.yaml@v4.0.0
+    uses: canonical/data-platform-workflows/.github/workflows/build_charm_without_cache.yaml@v4.1.0
 
   release:
     name: Release charm
     needs:
       - ci-tests
       - build
-    uses: canonical/data-platform-workflows/.github/workflows/release_charm.yaml@v4.0.0
+    uses: canonical/data-platform-workflows/.github/workflows/release_charm.yaml@v4.1.0
     with:
       channel: 8.0/edge
       artifact-name: ${{ needs.build.outputs.artifact-name }}
--- a/.github/workflows/sync_issue_to_jira.yaml
+++ b/.github/workflows/sync_issue_to_jira.yaml
@@ -9,7 +9,7 @@ on:
 jobs:
   sync:
     name: Sync GitHub issue to Jira
-    uses: canonical/data-platform-workflows/.github/workflows/sync_issue_to_jira.yaml@v4.0.0
+    uses: canonical/data-platform-workflows/.github/workflows/sync_issue_to_jira.yaml@v4.1.0
     with:
       jira-base-url: https://warthogs.atlassian.net
       jira-project-key: DPE
--- a/poetry.lock
+++ b/poetry.lock
@@ -138,17 +138,17 @@ uvloop = ["uvloop (>=0.15.2)"]
 
 [[package]]
 name = "boto3"
-version = "1.28.14"
+version = "1.28.15"
 description = "The AWS SDK for Python"
 optional = false
 python-versions = ">= 3.7"
 files = [
-    {file = "boto3-1.28.14-py3-none-any.whl", hash = "sha256:d06b8f8c962f320fc462503e476c5f5cffd6c9da5e090ca6cc7e9f0e931b2945"},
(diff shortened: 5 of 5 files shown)


taurus-forever (dismissed):

.github/workflows/ci.yaml
```diff
@@ -27,9 +27,10 @@ jobs:
         uses: runforesight/foresight-workflow-kit-action@v1
       - name: Checkout
         uses: actions/checkout@v3
-      - name: Install tox
-        # TODO: Consider replacing with custom image on self-hosted runner OR pinning version
-        run: python3 -m pip install tox
+      - name: Install tox & poetry
+        run: |
+          pipx install tox
```
paulomach, line 32: where in the workflow are we installing `pipx`?

pyproject.toml
```diff
@@ -1,7 +1,55 @@
...
+black = "^23.3.0"
+isort = "^5.12.0"
+
+[tool.poetry.group.lint]
+optional = true
+
+[tool.poetry.group.lint.dependencies]
+black = "^23.3.0"
+isort = "^5.12.0"
+flake8 = "^6.0.0"
```
paulomach, line 33: since we are refactoring this, wouldn't be nice time to move to `ruff`

tox.ini
```diff
@@ -14,61 +13,51 @@ all_path = {[vars]src_path} {[vars]tests_path} {[vars]lib_path}
...
-    flake8-builtins
-    pyproject-flake8
-    pep8-naming
-    isort
-    codespell
-    shellcheck-py==0.9.0.2
-allowlist_externals = /bin/bash
+allowlist_externals =
+    {[testenv]allowlist_externals}
+    /bin/bash
```
paulomach, line 38: bash is only used on L49? Why not just add `find` instead

requirements.txt
```diff
@@ -1,8 +1,199 @@
-boto3==1.26.113
-cosl==0.0.5
-cryptography==40.0.1
-jsonschema==4.17.3
-ops >= 2.0.0
-pyOpenSSL >= "23.0.0"
-tenacity==8.2.2
-urllib3==1.26.15
```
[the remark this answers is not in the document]
paulomach: not sure about pyOpenSSL

urllib3 is used on integration test for test metric endpoint, [here](https://github.com/canonical/mysql-operator/blob/1d662df9200a28e167f633a99772bffde45b6cea/tests/integration/high_availability/test_replication.py#L52)

paulomach (approved):
Optional comments and questions. lgtm

requirements.txt
```diff
@@ -1,9 +0,0 @@
-boto3==1.26.141
-cosl==0.0.5
-cryptography==40.0.2
-jsonschema==4.17.3
-ops >= 2.0.0
-pydantic==1.10.8
-pyOpenSSL >= "23.0.0"
-tenacity==8.2.2
-urllib3==1.26.16
```
taurus-forever, line 9: Can you please sync with @dragomirp here.
I would like to keep the common approach for both MySQL and PostgreSQL charms/sources, see
https://github.com/canonical/postgresql-operator/blob/main/requirements.txt

pyproject.toml
taurus-forever: I would not mix those topics:
1) migration to `poetry`
2) add `renovate`
3) `ruff` lint

Step-by-step... we are in step 1 here. :-D

taurus-forever (dismissed):
LGTM, but sync with Dragomir for the common approach. Tnx!

shayancanonical (approved):

pyproject.toml
```diff
@@ -1,7 +1,66 @@
...
+name = "charm"
+version = "0.1.0"
+description = ""
+authors = []
+
+[tool.poetry.dependencies]
+python = "^3.10"
+ops = "^2.4.1"
+tenacity = "^8.2.2"
+boto3 = "^1.28.7"
```
dragomirp, line 15: Caret pins mean less renovate PRs, but there's no real guarantee that minor version changes won't break things. We'll have to pick our poison here, either have to vet more PRs or accept that dependencies can break.

pyproject.toml
```diff
@@ -1,7 +1,66 @@
...
+
+[tool.poetry.group.charm-libs.dependencies]
+# data_platform_libs/v0/data_interfaces.py
+ops = ">=2.0.0"
+# grafana_agent/v0/cos_agent.py
+cosl = "*"
+pydantic = "<2"
+# tls_certificates_interface/v1/tls_certificates.py
+cryptography = "*"
+jsonschema = "*"
```
dragomirp, line 27: Is this group manually maintained? If so, I think people will only update it when things break, so I don't think it will warn us of dependency conflicts.

requirements.txt
dragomirp, line 1: FWIW, using renovate's `pip_requirements` manager worked fine with persistent  `requirements.txt`. Keeping it around gives us another manual escape hatch for issues like the libjuju's pin for pyyaml. IMHO it would be optimal to export on renovate PRs or when updating the lock.

pyproject.toml
dragomirp: If so, why not just pin the exact versions?  That way local and CI builds will all pull a known working version and we wait for renovate to move the pin. Otherwise bulids happening before the renovate schedule will still potentially face issues.

requirements.txt
dragomirp: > AFAIK, renovate's `pip_requirements` manager has no knowledge of poetry and should not be used if poetry is used—poetry.lock is useless if we have a separate requirements.txt file that bypasses it

It worked fine (YMMV) in the Postgresql charms, since it's mostly bumping versions. When it breaks, it's a matter of either pinning a version in renovate.json or reexporting from poetry.
 
> Poetry itself has the tools to resolve issues (as long as you use version ranges instead of exact versions in pyproject.toml). poetry.lock and, by extension, requirements.txt should not be edited manually—poetry does the dependency resolution for you

I don't think Poetry alone would have solved the pyyaml breakage we just had. It resolved the correct version, but that version was no longer working. If something similar happens, we'll either have to stick to our guns and wait for an upstream fix or readd `requirements.txt` and cleanup after the upstream fix.

shayancanonical (approved):
lgtm

paulomach (approved):

paulomach (dismissed):
sensible

charmcraft.yaml
```diff
@@ -14,6 +14,13 @@ bases:
         channel: "22.04"
 parts:
   charm:
+    override-pull: |
+      craftctl default
+      if [[ ! -f requirements.txt ]]
+      then
+          echo 'ERROR: Use "tox run -e build" instead of calling "charmcraft pack" directly'
+          exit 1
+      fi
```
taurus-forever, line 23: Just to follow Shell best practices:
1) print errors to STDERR
2) nit: consider to use [Google Shell code style](https://google.github.io/styleguide/shellguide.html#environment)

```
      craftctl default
      if [[ ! -f requirements.txt ]]; then
          echo 'ERROR: Use "tox run -e build" instead of calling "charmcraft pack" directly' >&2
          exit 1
      fi
```

CONTRIBUTING.md
```diff
@@ -20,12 +20,18 @@ this operator.
...
-You can create an environment for development with `tox`:
+You can create an environment for development:
 
 ```shell
-tox devenv -e integration
-source venv/bin/activate
+poetry install
 ```
 
 ### Testing
```
taurus-forever, line 37: Did you left "charmcraft pack" by intention in line 52?
> tox run -e build   ?

requirements.txt
taurus-forever: @carlcsaposs-canonical please show example of addressing recent PyYAML issue using poetry:
* example of commands to call
* poetry.lock result readability / maintainability
Tnx!

taurus-forever (commented):
I like it! a couple of questions.

paulomach (approved):
looks good, left two non-blocking questions

taurus-forever (approved):
Nice progress here! Let's give it a try.
Please wait Paulo review before submitting.

PR #207, opened 2023-05-08
Delete users created by MySQL Router charm when relation broken

carlcsaposs-canonical:
Ported from https://github.com/canonical/mysql-k8s-operator/pull/220

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -90,7 +90,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 25
+LIBPATCH = 26
 
 UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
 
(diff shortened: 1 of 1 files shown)


shayancanonical (commented):
I believe we can consolidate some of the methods in the mysql charm lib in this PR to result in cleaner code.

The way I was thinking about it, we would have:
1. one method (`create_application_database_and_scoped_user`) which would take in an arbitrary label and label value that it applies to the created user. In the case of mysql_provider, this would be `relation_id`. In the case of shared_db, it would be `unit_name` and in the case of mysql_router, it would be `mysql_relation_id`
2. we consolidate the `delete_users_for_unit` and `delete_users_for_relation` into one method: `delete_users_with_label`. We then appropriately pass in the label and label value to delete matched users in the method

I like the approach with prepared statements, as it reduces the chances of a race condition where there is a user created between the time we query for users and execute delete queried users in `delete_users_for_unit.

Additionally, I just noticed that `delete_users_for_relation` does not check if the `primary_addressed` returned is `None` while `delete_users_for_unit` does -> the consolidation of the methods may help eliminate this issue

shayancanonical:
re: pushing the consolidation to a future date - i am ok with it, but by my gut feeling is that the consolidation work is not too large. if it were me, i would timebox the effort to an afternoon. however, i'll defer to the opinions of @paulomach or @taurus-forever

lib/charms/mysql/v0/mysql.py
```diff
@@ -459,7 +460,14 @@ def configure_mysqlrouter_user(
             raise MySQLConfigureRouterUserError(e.message)
 
     def create_application_database_and_scoped_user(
-        self, database_name: str, username: str, password: str, hostname: str, unit_name: str
+        self,
+        database_name: str,
+        username: str,
+        password: str,
+        hostname: str,
+        *,
```
paulomach, line 468: nice touch

paulomach (commented):
Looks good, but I'll second Shayan - It's not too much effort to round that edges, and you can count on help for testing

delgod (approved):

delgod (approved):

shayancanonical (approved):

paulomach (approved):
awesome

PR #208, opened 2023-05-10
Cleanup router metadata

carlcsaposs-canonical:
Depends on #207 (change base branch after #207 is merged)

Ported from https://github.com/canonical/mysql-k8s-operator/pull/224

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -632,7 +632,7 @@ class MySQLBase(ABC):
             raise MySQLDeleteUserError("Unable to query cluster primary address")
         drop_user_command = [
             f"shell.connect('{self.server_config_user}:{self.server_config_password}@{primary_address}')",
-            f'session.run_sql("DROP USER `{username}`")',
+            f"session.run_sql(\"DROP USER `{username}`@'%'\")",
         ]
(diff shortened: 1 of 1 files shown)


src/relations/mysql_provider.py
```diff
@@ -256,3 +264,34 @@ def _on_database_broken(self, event: RelationBrokenEvent) -> None:
...
+        if event.departing_unit.app.name == self.charm.app.name:
+            return
+
+        users = self.charm._mysql.get_mysql_router_users_for_unit(
+            relation_id=event.relation.id, mysql_router_unit_name=event.departing_unit.name
+        )
+        if not users:
+            return
+
+        if len(users) > 1:
```
shayancanonical, line 281: couldnt there be more than 1 applications related to the router? if so, wouldnt there be more than 1 user returned?

shayancanonical: > The unit name contains the app name

if that's the case, should we rename `created_by_juju_unit` to `created_by_router_identifier` where `router_identifier` is a combination of the juju model, application and unit for the router?

paulomach (approved):
lgtm

shayancanonical (approved):

delgod (approved):

PR #210, opened 2023-05-15
Skip database creation for `mysqlrouter` extra user role

carlcsaposs-canonical:
Ported from https://github.com/canonical/mysql-k8s-operator/pull/227

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -65,6 +65,7 @@ error handling on the subclass and in the charm code.
 
 """
 
+import dataclasses
 import json
 import logging
 import re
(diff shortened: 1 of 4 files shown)


delgod (approved):

lib/charms/mysql/v0/mysql.py
```diff
@@ -90,7 +90,7 @@ def wait_until_mysql_connection(self) -> None:
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 26
+LIBPATCH = 28
```
shayancanonical, line 93: should this libpatch be 27 instead?

shayancanonical (approved):

src/relations/mysql_provider.py
```diff
@@ -212,16 +212,25 @@ def _on_database_requested(self, event: DatabaseRequestedEvent):
...
+                    "%",
+                    # MySQL Router charm does not need a new database
+                    create_database=False,
+                )
                 self.charm._mysql.grant_privileges_to_user(
                     db_user, "%", ["ALL PRIVILEGES"], with_grant_option=True
                 )
+            else:
+                # TODO:
+                # add setup of tls, tls_ca and status
```
taurus-forever, line 230: JFMI, is it still topical? mysql-router-k8s supports TLS since the last sprint AFAIK, mysql-router is subordinate == no TLS.

lib/charms/mysql/v0/mysql.py
taurus-forever: I suspect CI will fail if you try to release LIBPATCH = 27 after merging this LIBPATCH = 28 (also, AFAIK charmcraft will refuse to upload the library with gap in LIBPATCH). I could be wrong.

taurus-forever (approved):
LGTM, one question.

PR #219, opened 2023-05-30
DPE-1688 DPE-1913 Add alert logs for success/failure of backups + change temp backup dir

shayancanonical:
## Issue
1. We are not logging the success or failure of both the `create-backup` and `restore` actions
2. We are currently creating a temp backup and restore directories in the mysql data directory (to ensure that the temp backup directory is on the same volume as the data directory to reduce IOPS latency). However, we expanded the volume to include the snap common directory, so the temp directories no longer need to be in the data directory.
 
## Solution
1. Log failures (as warnings) and successes (as info) to the debug log. The failures have prefix `Backup failed`/`Restore failed`, the successes have prefix `Backup successful`/`Restore successful`
2. Move the temp directories out of the data directory, but within the snap common directory (where the volume is mounted). Additionally, make sure that the backups are backwards compatible (by still removing any temp directories in the data directory from old backups)

### DIFF
--- a/lib/charms/mysql/v0/backups.py
+++ b/lib/charms/mysql/v0/backups.py
@@ -227,12 +227,12 @@ Stderr:
         logger.info("A backup has been requested on unit")
 
         if not self.charm.s3_integrator_relation_exists:
-            logger.warning("Backup failed: missing relation with S3 integrator charm")
+            logger.error("Backup failed: missing relation with S3 integrator charm")
             event.fail("Missing relation with S3 integrator charm")
             return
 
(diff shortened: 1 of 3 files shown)


taurus-forever (changes requested):
I second to Carl. Why errors for failed action showed as warnings?

shayancanonical:
Converted warning message for failures to `error` in 94d7e01d39aa7b5376abaea44f6cb87fe9e84ba3

taurus-forever (approved):

PR #220, opened 2023-05-31
Add memory constraints in conftest and change unit logging to debug in test models

shayancanonical:
## Issue
1. We are not setting any constraints on the mysql units (which results in mysql setting `innodb_buffer_pool_size` from calculations of the entire host machine). This may lead to some failures (OOM) in integration tests
2. We are not printing out debug log in the integration tests, making it very difficult to see what was happening when a test failed

## Solution
1. Set a memory constraint of 1G
2. Print unit debug logs in the output

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -3,8 +3,8 @@
 import dataclasses
 import json
 import os
-from pathlib import Path
 import subprocess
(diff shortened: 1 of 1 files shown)


shayancanonical:
Closing, due to the introduction of a `testing` profile in a separate PR

PR #223, opened 2023-06-02
add auto-tuning for max_connections

delgod:
## Issue

MySQL `max_connections` is neither being tunned nor being exposed to configuration, staying at the default 151 connections.

## Solution

Add sane auto-tune strategy while proper configuration management is not implemented.

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -1458,12 +1458,11 @@ class MySQLBase(ABC):
             raise MySQLGetAutoTunningParametersError("Error computing buffer pool parameters")
 
     def get_max_connections(self) -> int:
-        """Calculate max_connections parameter for the instance.
-        """
+        """Calculate max_connections parameter for the instance."""
         # Reference: based off xtradb-cluster-operator
         # https://github.com/percona/percona-xtradb-cluster-operator/blob/main/pkg/pxc/app/config/autotune.go#L61-L70
 
-        bytes_per_connection = 12582912 # 12 Megabytes
+        bytes_per_connection = 12582912  # 12 Megabytes
         total_memory = None
 
         try:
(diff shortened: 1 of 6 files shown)


lib/charms/mysql/v0/mysql.py
```diff
@@ -1513,9 +1514,29 @@ def get_innodb_buffer_pool_parameters(self) -> Tuple[int, Optional[int]]:
...
+        except Exception:
+            logger.exception("Failed to compute innodb buffer pool parameters")
+            raise MySQLGetAutoTunningParametersError("Error computing buffer pool parameters")
+
+    def get_max_connections(self) -> int:
+        """Calculate max_connections parameter for the instance."""
+        # Reference: based off xtradb-cluster-operator
+        # https://github.com/percona/percona-xtradb-cluster-operator/blob/main/pkg/pxc/app/config/autotune.go#L61-L70
+
+        bytes_per_connection = 12582912  # 12 Megabytes
```
paulomach, line 1526: I wonder if this is way too conservative and can still not be enough.

taurus-forever (approved):
Just for the history: this will decrease the current max_connections in CI tests.

We allowed 111 connections in the past, but currently we constraint RAM to 1GB, which means ~85 max connections allowed (42? connections for RAM=500Mb installations).

taurus-forever (approved):
Just for the history: this will decrease the current max_connections in CI tests.

We allowed 111 connections in the past, but currently we constraint RAM to 1GB, which means ~85 max connections allowed (42? connections for RAM=500Mb installations).

lib/charms/mysql/v0/mysql.py
delgod: Absolutely conservative, it is sort of "safe defaults".
The idea is to provide safe default, and if the user knows the biggest value that does not produce OOM for his particular workload, then he will customize the value via "juju config" (when we have config management).

PR #224, opened 2023-06-02
Optimize mysqlsh calls on library

paulomach:
## Issue

Methods relying on retrieving the cluster primary are doing two or more calls to mysqlsh either through local Popen (vm) or pebble (k8s) with extra overhead.

## Solution

* use `connect_to_primary` mysqlsh api call that takes care of switching the session to the primary
* consolidate calls that made unnecessary multiple shell calls

_chore_: fixes for unit tests warning removals


taurus-forever (approved):
Tnx!

PR #229, opened 2023-06-16
DPE-2124 More memory saves

paulomach:
## Issue

Small machines (_I'm looking at you standard github runner_) may suffer with not enough memory.

## Solution

Disabling some of the memory instruments helps save ~100MB of memory on mysqld.
[Ref. link.](https://dev.mysql.com/doc/mysql-perfschema-excerpt/8.0/en/performance-schema-memory-summary-tables.html)


paulomach:
> Didn't we consider to go with "profile=devel"?

Yes! this was before the decision was made, I'll just close it.

PR #230, opened 2023-06-17
Basic prometheus alert rule

paulomach:
Same as k8s [PR244](https://github.com/canonical/mysql-k8s-operator/pull/244)


taurus-forever (approved):
Nice check :-)

PR #232, opened 2023-06-21
Handle IP address changes of units

shayancanonical:
## Issue
We are currently not handling the IP address change of a unit in the cluster.

## Solution
Handle the IP address change of a unit (whether in a cluster or a single node).

Steps to recover involve:
1. Remove plugin `group_replication`
2. Update `report_host` in the custom mysqld config file and restart mysqld
3. Add node back to cluster (or recreate cluster if there is only 1 unit in the cluster)

If the transaction set in the unit is a subset of the transaction set in the cluster, then upon addition into the cluster, distributed recovery will be in effect. Else, the data will be cloned from one of the units in the cluster.

## Considerations
- How will scaling down to 1 unit (preserving the data volume), recovering from IP address change, and then scaling up effect SST (because the GTID may be different after re-creating a cluster in the recovery process)?

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -77,6 +77,7 @@ from tenacity import (
     retry,
     retry_if_exception_type,
     stop_after_attempt,
+    stop_after_delay,
     wait_fixed,
     wait_random,
 )
@@ -91,7 +92,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 34
+LIBPATCH = 35
 
 UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
 UNIT_ADD_LOCKNAME = "unit-add"
@@ -297,6 +298,22 @@ class MySQLLockAcquisitionError(Error):
 
 class MySQLRescanClusterError(Error):
     """Exception raised when there is an issue rescanning the cluster."""
+
+
+class MySQLUninstallGroupReplicationError(Error):
+    """Exception raised when there is an issue uninstalling group replication."""
+
+
+class MySQLGetVariableValueError(Error):
+    """Exception raised when there is an issue getting variable value."""
+
+
+class MySQLDropMetadataSchemaError(Error):
+    """Exception raised when there is an issue dropping the metadata schema."""
+
+
+class MySQLSuperReadOnlyEnabledError(Error):
+    """Exception raised when the super_read_only variable is eneabled."""
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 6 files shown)


lib/charms/mysql/v0/mysql.py
```diff
@@ -77,6 +77,7 @@ def wait_until_mysql_connection(self) -> None:
     retry,
     retry_if_exception_type,
     stop_after_attempt,
+    stop_after_delay,
     wait_fixed,
     wait_random,
 )
```
shayancanonical, line 83: TODO: update the lib patch version

lib/charms/mysql/v0/mysql.py
```diff
@@ -1891,6 +1908,85 @@ def delete_temp_restore_directory(
...
+        """
+        logger.info("Waiting until MySQL super_read_only is disabled")
+        if self.get_variable_value("super_read_only") == "ON":
+            logger.info("MySQL super_read_only enabled. Retrying in 5 seconds")
+            raise MySQLSuperReadOnlyEnabledError("MySQL super_read_only enabled")
+        logger.info("MySQL super_read_only disabled")
+
+    def uninstall_group_replication(self) -> None:
+        """Uninstalls group replication plugin after setting super_read_only off."""
+        disable_super_read_only_command = "SET GLOBAL super_read_only = 0"
```
delgod, line 1925: Just Q: why do we need to reset super_read_only before disabling GR?

src/charm.py
```diff
@@ -314,6 +322,106 @@ def _handle_non_online_instance_status(self, state) -> None:
...
+            return True
+
+        try:
+            report_host = self._mysql.get_variable_value("report_host")
+        except MySQLGetVariableValueError:
+            report_host = None
+
+        try:
+            hostname = socket.gethostname()
+            unit_ip = socket.gethostbyname(hostname)
```
delgod, line 343: I think it is possible to get `127.0.0.1` here under some `/etc/hosts` configuration.
maybe https://stackoverflow.com/a/28950776

src/charm.py
```diff
@@ -314,6 +322,106 @@ def _handle_non_online_instance_status(self, state) -> None:
...
+
+            if report_host and report_host != unit_ip:
+                logger.info(
+                    "Restarting mysqld, since it was started with --super-read-only option"
```
delgod, line 347: But why do we need to restart? the read-only mode can be removed without restart

delgod (changes requested):

lib/charms/mysql/v0/mysql.py
```diff
@@ -1891,6 +1908,85 @@ def delete_temp_restore_directory(
...
+                user=self.server_config_user,
+                password=self.server_config_password,
+            )
+        except MySQLClientError as e:
+            logger.exception("Failed to uninstall plugin group replication")
+            raise MySQLUninstallGroupReplicationError(e.message)
+
+    def get_variable_value(self, variable_name: str) -> None:
+        """Retrieves the value of the provided variable from MySQL."""
+        retrieve_variable_command = f"SHOW VARIABLES LIKE '{variable_name}'"
```
paulomach, line 1959:
```suggestion
        retrieve_variable_command = f"SHOW VARIABLES WHERE VARIABLE_NAME '{variable_name}'"
```
or
```suggestion
        retrieve_variable_command = f"SELECT @@`{variable_name}`"
```

taurus-forever:
@shayancanonical as discussed, it smells a bit. After discussion with @delgod , let's try the second approach (keep cluster configured on DNS names, but add manual /etc/hosts handling (like legacy charm did).
Let's compare the pros/cons for both cases before merging this approach. Thank you!

shayancanonical:
Closing in lieu of the implementation in https://github.com/canonical/mysql-operator/pull/237 (which is far cleaner and modular)

PR #249, opened 2023-07-24
Use snap with ppa sources + other misc snap related improvements

shayancanonical:


src/constants.py
```diff
@@ -32,7 +32,7 @@
 TLS_SSL_CERT_FILE = "custom-server-cert.pem"
 MYSQL_EXPORTER_PORT = "9104"
 CHARMED_MYSQL_SNAP_NAME = "charmed-mysql"
-CHARMED_MYSQL_SNAP_REVISION = 51
+CHARMED_MYSQL_SNAP_REVISION = 65
```
shayancanonical, line 35: TODO: update revision once new revision of snap is published to 8.0/edge once https://github.com/canonical/charmed-mysql-snap/pull/35 is merged

src/mysql_vm_helpers.py
```diff
@@ -561,11 +573,13 @@ def _run_mysqlsh_script(self, script: str, timeout=None) -> str:
...
             _file.flush()
 
-            # Specify python as this is not the default in the deb version
-            # of the mysql-shell snap
             command = [CHARMED_MYSQLSH, "--no-wizard", "--python", "-f", _file.name]
 
             try:
+                # need to change permissions since charmed-mysql.mysqlsh runs as
+                # snap_daemon
+                subprocess.run(["chmod", "666", _file.name], check=True)
```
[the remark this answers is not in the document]
shayancanonical: i dont think it makes a big difference, as the file is a named temporary file. i can chown it instead since the file may contain passwords locally

tests/integration/high_availability/test_replication.py
```diff
@@ -53,6 +56,50 @@ async def test_exporter_endpoints(ops_test: OpsTest, mysql_charm_series: str) ->
...
+            "set",
+            "charmed-mysql",
+            f"exporter.password={monitoring_credentials['password']}",
+        )
+        assert return_code == 0
+
+        return_code, _, _ = await ops_test.juju(
+            "ssh", unit.name, "sudo", "snap", "start", "charmed-mysql.mysqld-exporter"
+        )
+        assert return_code == 0
```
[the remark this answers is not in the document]
shayancanonical: because we are not only starting the exporter service when a relation with cos is formed. by default, the service is disabled. it would be too much to set up a cos bundle to test the exporter, so this is a lighter version to make sure that the exporter service still works

src/mysql_vm_helpers.py
```diff
@@ -539,11 +541,21 @@ def connect_mysql_exporter(self) -> None:
...
             logger.exception("An exception occurred when setting up mysqld-exporter.")
-            raise MySQLExporterConnectError
+            raise MySQLExporterConnectError("Error setting up mysqld-exporter")
+
+    def stop_mysql_exporter(self) -> None:
+        """Stop the mysqld exporter."""
+        try:
+            snap_service_operation(
+                CHARMED_MYSQL_SNAP_NAME, CHARMED_MYSQLD_EXPORTER_SERVICE, "stop"
+            )
```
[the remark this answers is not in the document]
shayancanonical: enabling and disabling upon starting and stopping exporter service respectively in e5d053cbdbe07c4a3fd9a80109c79de8c4fa58c5

src/mysql_vm_helpers.py
```diff
@@ -755,7 +771,7 @@ def snap_service_operation(snapname: str, service: str, operation: str, enable=F
             selected_snap.start(services=[service], enable=enable)
             return selected_snap.services[service]["active"]
         else:
-            selected_snap.stop(services=[service])
+            selected_snap.stop(services=[service], disable=disable)
```
delgod, line 774: feel free to ignore it: why not?
```suggestion
            selected_snap.stop(services=[service], disable=not enable)
```

src/charm.py
```diff
@@ -381,6 +387,35 @@ def _on_update_status(self, _) -> None:
...
+        """
+        if not self._is_peer_data_set:
+            return
+
+        try:
+            self._mysql.stop_mysql_exporter()
+        except MySQLExporterConnectError:
+            logger.exception("Unable to stop the mysqld-exporter snap service")
+            self.unit.status = BlockedStatus("Unable to stop mysqld-exporter")
+
```
taurus-forever, line 418: IMHO.
1) add debug log when you are deferring an event (line 396). It is very hard to guess in production why things didn't happen.
2) IMHO, we should rise Error and not Blocked if we failed to start/stop. User will be able to call "juju resolve", otherwise charm will stuck in blocked state if it failed to stop mysqld-exporter. Isn't it?

taurus-forever (approved):
LGTM, two comments which changes nothing but UX.

src/charm.py
paulomach: I 2nd point 2, specially because the exporter is a not the main charm service

paulomach (approved):
lgtm

src/mysql_vm_helpers.py
shayancanonical: If we do `not enable`, we would always need to specify enable if we would like to avoid disabling the service when stopping it, which seems a bit counter intuitive

e.g.
```
snap_service_operation(snap_name, service_name, "stop", enable=True)
``

File: src/mysql_vm_helpers.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (shayancanonical) [reply src/mysql_vm_helpers.py]: i am not sure there is such a use case at the moment. i would not want a use case to arise in the future when the enabling/disabling is done transparently, which is why i added the kwargs

File: src/constants.py (hunk 1 above)
Comment (shayancanonical) [reply src/constants.py]: updated to latest 8.0/edge (revision 66) in 9f7d5c199f5e6824c0f44babf3c7917c58930592

File: src/mysql_vm_helpers.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (shayancanonical) [reply src/mysql_vm_helpers.py:555]: good point, always enable when starting and disable when stopping in 9f7d5c199f5e6824c0f44babf3c7917c58930592

File: src/mysql_vm_helpers.py [hunk 7]
```diff
@@ -561,11 +573,13 @@ def _run_mysqlsh_script(self, script: str, timeout=None) -> str:
...
             _file.flush()
 
-            # Specify python as this is not the default in the deb version
-            # of the mysql-shell snap
             command = [CHARMED_MYSQLSH, "--no-wizard", "--python", "-f", _file.name]
 
             try:
+                # need to change permissions since charmed-mysql.mysqlsh runs as
+                # snap_daemon
+                subprocess.run(["chown", "snap_daemon:root", _file.name], check=True)
```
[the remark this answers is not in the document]
shayancanonical: addressed in 9f7d5c199f5e6824c0f44babf3c7917c58930592

src/mysql_vm_helpers.py
shayancanonical: removed the `enable` and `disable` kwargs in 9f7d5c199f5e6824c0f44babf3c7917c58930592 (and always enable upon start and always disable upon stop)

src/charm.py
shayancanonical: addressed in 9f7d5c199f5e6824c0f44babf3c7917c58930592

tests/integration/high_availability/test_replication.py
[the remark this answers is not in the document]
paulomach: IMO I don't think it's that much of a hassle, since you won't need to setup a full cos deployment, but just a relation with the subordinated `grafana-agent` charm, which is fairly light weight

paulomach (approved):
lgtm - i left a non blocking comment though

paulomach (approved):

PR #253, opened 2023-07-28
port from k8s and fixes for constrained memory

paulomach:
## Issue

Backport changes from k8s [PR#256](https://github.com/canonical/mysql-k8s-operator/pull/256)


lib/charms/mysql/v0/backups.py
```diff
@@ -98,10 +99,14 @@ def is_unit_blocked(self) -> bool:
 LIBPATCH = 6
```
shayancanonical, line 99: would we need to update the libpatch?

paulomach: Given that there's no functional changes (just typing) it does not seems right

shayancanonical: i suggested the libpatch update so that we can easily port improvements over to the k8s charm (else these changes may get lost at some point)

shayancanonical (dismissed):

shayancanonical (approved):

PR #261, opened 2023-08-03
DPE-2215 Fix wait timeout for shared-db

paulomach:
## Issue

Mismatch between relation data and server sysvar `wait_timeout`

## Solution

Use mysqld default `wait_timeout` value.


shayancanonical (approved):

paulomach:
> @paulomach Does this apply for mysql-router VM as well?

No, it's already with the [correct value](https://github.com/canonical/mysql-router-operator/blob/a98b4918226cfda12b3f1832b9943ba3de609583/src/relations/deprecated_shared_db_database_provides.py#L102).

PR #276, opened 2023-08-07
Add env to current env instead of overriding env in subprocess.run

carlcsaposs-canonical:
Fixes backup integration tests on self-hosted runners

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -420,8 +420,9 @@ class MySQLCharmBase(CharmBase):
             return
 
         new_password = event.params.get("password") or generate_random_password(PASSWORD_LENGTH)
+        host = "%" if username != ROOT_USERNAME else "localhost"
 
-        self._mysql.update_user_password(username, new_password)
+        self._mysql.update_user_password(username, new_password, host=host)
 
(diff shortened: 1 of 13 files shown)


src/mysql_vm_helpers.py
```diff
@@ -455,12 +455,15 @@ def _execute_commands(
...
             user: the user with which to execute the commands
             group: the group with which to execute the commands
-            env: the environment variables to execute the commands with
+            env_extra: the environment variables to add to the current process’ environment
 
         Returns: tuple of (stdout, stderr)
 
         Raises: MySQLExecError if there was an error executing the commands
         """
+        env = os.environ.copy()
```
shayancanonical, line 464: i wonder if there are env vars like `$HOME` or `$USER` which we should not be passing to subprocess (as they may result in permissions issues)

shayancanonical (approved):

taurus-forever (dismissed):
Tnx!

paulomach (dismissed):

shayancanonical (approved):

shayancanonical (approved):

paulomach (approved):

taurus-forever (approved):
Super!

PR #285, opened 2023-08-09
DPE-2352 Restart mysql exporter upon monitoring password change

shayancanonical:
## Issue
We are currently not restarting the mysqld exporter when the monitoring user password changes. 

## Solution
Restart the mysqld exporter (stop and connect again with the new password) when the password changes

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -88,6 +88,7 @@ from constants import (
     BACKUPS_USERNAME,
     CLUSTER_ADMIN_PASSWORD_KEY,
     CLUSTER_ADMIN_USERNAME,
+    COS_AGENT_RELATION_NAME,
     MONITORING_PASSWORD_KEY,
     MONITORING_USERNAME,
     PASSWORD_LENGTH,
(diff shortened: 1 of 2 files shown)


lib/charms/mysql/v0/mysql.py
```diff
@@ -404,6 +405,12 @@ def _on_set_password(self, event: ActionEvent) -> None:
 
         self.set_secret("app", secret_key, new_password)
 
+        if (
+            username == MONITORING_USERNAME
+            and len(self.model.relations.get(COS_AGENT_RELATION_NAME, [])) > 0
+        ):
+            self._mysql.restart_mysql_exporter()
```
paulomach, line 412: This should be also be added as an abstract method on the `MySQLBase` to enforce k8s implementation

shayancanonical: Addressed in [116d61f](https://github.com/canonical/mysql-operator/pull/285/commits/116d61fcb0bc433487fe20382c450f4ab67b2c20)

taurus-forever (approved):
Thank you for taking care here! Nice!

PR #289, opened 2023-08-11
Add `has_cos_relation` to the MySQLCharmBase class

shayancanonical:
## Issue
- We added a `has_cos_relation` property in a branch in the k8s charm. It seems like this property would be good to have in the vm charm as well as the k8s charm
- Sometimes, we pass the current_event (event of a handler) down a chain of methods before it gets used

## Solution
- Add `has_cos_relation` property to the MySQLBaseCharm class
- Introduce a `self.current_event` field (initialized to `None`) that can be set on the charm class by an event and then used in helper methods

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -111,7 +111,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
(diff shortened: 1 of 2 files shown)


paulomach (approved):
lgtm

marceloneppel (approved):
LGTM! Thanks for the details on MM.

shayancanonical:
Thanks for the reviews!

PR #291, opened 2023-08-11
Add juju agent version requirement to metadata.yaml

carlcsaposs-canonical:
Ported from https://github.com/canonical/mysql-k8s-operator/pull/277

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -56,4 +56,10 @@ storage:
     location: /var/snap/charmed-mysql/common
 
 assumes:
-  - juju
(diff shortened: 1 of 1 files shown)


taurus-forever (changes requested):
Do we enforce 2.9.44 for VM too? Pebble is on K8s only. Nope?

paulomach:
> Do we enforce 2.9.44 for VM too? Pebble is on K8s only. Nope?

@carlcsaposs-canonical , Alex is correct here, the constraint does not apply for VM

PR #292, opened 2023-08-11
DPE-1800 in place upgrades

paulomach:
### In place upgrades for mysql vm

Usage upgrade lib and follow procedure per spec.

## Solution

Implemented upgrade module, unit test and simple case integration test.
To follow:
* rollback testing
* test with patched snap versions
* discourse doc page

### Manual Testing (with same local built charm)

Instructions (happy path):

```shell

git clone git@github.com:canonical/mysql-operator.git
cd mysql-operator
git checkout feature/dpe-1800-in-place-upgrade

charmcraft pack
# deploy local charm
juju deploy ./*.charm -n 3

# After units settle, run pre-upgrade-check to build update-stack.
# This will change primary to the leader if not already
juju run-action mysql/leader pre-upgrade-check --wait

# inspect update relation data
juju show-unit mysql/0|yq '.*.relation-info[]|select(.endpoint=="upgrade")'

# refresh command - we are updating to the same charm, which will pass version checking
# optionally, patch the `src/constants.py`, bump snap revision (66->67) and pack the charm
juju refresh --path ./*.charm mysql

# enjoy the process

```

### DIFF
--- a/src/dependency.json
+++ b/src/dependency.json
@@ -1,0 +1,14 @@
+{
+  "charm": {
+    "dependencies": {},
+    "name": "mysql",
+    "upgrade_supported": ">0",
+    "version": "1"
+  },
+  "snap": {
+    "dependencies": {},
+    "name": "charmed-mysql",
+    "upgrade_supported": "~8.0.31",
+    "version": "8.0.32"
+  }
+}
--- a/actions.yaml
+++ b/actions.yaml
@@ -49,3 +49,7 @@ restore:
     backup-id:
       type: string
       description: A backup-id to identify the backup to restore (format = %Y-%m-%dT%H:%M:%SZ)
+
+pre-upgrade-check:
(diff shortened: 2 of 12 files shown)


src/dependency.json
```diff
@@ -0,0 +1,14 @@
...
+    "dependencies": {},
+    "name": "mysql",
+    "upgrade_supported": ">0",
+    "version": "1"
+  },
+  "snap": {
+    "dependencies": {},
+    "name": "charmed-mysql",
+    "upgrade_supported": "~8.0.31",
+    "version": "8.0.32"
```
taurus-forever, line 12: the current version is 8.0.33 already

taurus-forever (approved):
Nice. Testing it locally right now, but lets merge edge to run more tests for different track.
Great work. Thank you!

shayancanonical (approved):
lgtm, but the minor nit that alex mentioned: mysql is already running v8.0.33

PR #306, opened 2023-08-23
Use 60s interval for fast forward instead of 5s

carlcsaposs-canonical:
When wait_for_idle called in fast forward

wait_for_idle requires all units to be idle for 15 seconds in a row. Fast forward triggers update status events every 5 seconds, which causes the charm to execute (not idle). This causes wait_for_idle to stall on less powerful hardware (e.g. `large` self-hosted runners) and potentially slows down integration tests on more powerful hardware


shayancanonical (commented):
should this change also be applied to all fast_forwards in the self healing test suite?

shayancanonical (approved):

paulomach (approved):

PR #310, opened 2023-08-25
DPE-2425 Update charmed-mysql snap to revision 69 (MySQL v8.0.34)

shayancanonical:
## Issue
We published a new snap (for v8.0.34)

## Solution
Update the charm to use this new snap


taurus-forever (approved):
2 Carl, I think we can merge it AS IS to rebase https://github.com/canonical/mysql-operator/pull/301 and test it using new mysql (as currently we are testing there without workload upgrade.

paulomach (approved):

PR #311, opened 2023-08-25
DPE-2214 Avoid resetting workload if recovery from unreachable state unsuccessful

shayancanonical:
## Issue
We are resetting the workload (wiping the data directory clean and re-initializing mysql) when a recovery (restart of `charmed-mysql` snap) from `unreachable` state is unsuccessful. This is super dangerous and we would like to avoid doing so 

## Solution
Instead, go to `BlockedStatus` to allow the admin to determine how to best resolve the issue

### DIFF
--- a/lib/charms/data_platform_libs/v0/upgrade.py
+++ b/lib/charms/data_platform_libs/v0/upgrade.py
@@ -284,7 +284,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 10
+LIBPATCH = 11
 
 PYDEPS = ["pydantic>=1.10,<2"]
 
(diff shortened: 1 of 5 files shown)


paulomach (approved):

paulomach (approved):

taurus-forever (approved):
1) I agree about the potential harm of the previous logic. LGTM.
2) I am not sure about the Blocked state here. What are the user steps to recover from Blocked state back to active here? Delete unit/re-add unit?

shayancanonical:
@taurus-forever about (2), I am not sure what the best status to put the unit in is. If we choose `ErrorStatus`, the hook will continue retrying and as a result forget about trying to scale it down. If `BlockedStatus`, the debug path is not simple for the admin -> the admin would have to paint a picture of the cluster status and determine how to best recover (or scale down if recovery is impossible)

PR #315, opened 2023-08-28
DPE-2401 Hold snap revision

paulomach:
## Issue

Snap not being held into revision.

## Solution

Hold it if not already held.


taurus-forever (approved):
LGTM.
Maybe we should also specify the proper channel (8.0/edge) in a moment of deploying the snap revision? Otherwise `snap list` output reports `latest/stable`, which is confusing.

PR #331, opened 2023-09-28
DPE-2654 profile-limits option

paulomach:
## Issue

* **main issue**: no fine tune for memory allocation
* configurations are being set without validation
* configurations changes are not triggering changes in the workload

## Solution

* Add `profile-limit-memory` config option
* Usage of data platform `data-models` lib for config validation
* Usage of `rollingOps` lib for workload restart orchestration, when the changed configs requires.
  * Add a method to defer primary to be last one to restart (mitigating primary switchover)

### Chore

* refactored `get_available_memory` to be simpler
* charm typing for `relations/mysql.py`
* typo fix
* usage of config values as properties, provided by `data-models` lib

PR #334, opened 2023-10-04
Avoid any actions when a unit is offline and allow mysql GR to recover

shayancanonical:
## Issue
The SST test fails because we are taking an action when a unit is offline (we are currently trying to rejoin the instance to the cluster).

## Solution
Do nothing, as MySQL Group Replication has logic around rejoining an offline instance.


paulomach (approved):
good - the upgrade from stable will probably keep failing - there's a fix on the PR #331

paulomach (approved):

PR #336, opened 2023-10-10
Avoid logging FLUSH LOG statements to the binlog

shayancanonical:
## Issue
By default, `FLUSH LOG` statements are logged to the binlog. When a member is not part of a group (self-healing scenarios), these log statements result in GTID conflicts which prevents the member from self-healing.

## Solution
Disable logging `FLUSH LOG` statements to the bin log.

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -116,7 +116,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 49
+LIBPATCH = 50
 
 UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
 UNIT_ADD_LOCKNAME = "unit-add"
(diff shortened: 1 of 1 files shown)


lib/charms/mysql/v0/mysql.py
```diff
@@ -2552,7 +2552,9 @@ def flush_mysql_logs(self, logs_type: MySQLTextLogs) -> None:
         """Flushes the specified logs_type logs."""
         flush_logs_commands = (
             f"shell.connect('{self.server_config_user}:{self.server_config_password}@{self.instance_address}')",
+            'session.run_sql("SET sql_log_bin = 0")',
```
paulomach, line 2555: does this has any effect on anything other then bin logs?

shayancanonical: https://dev.mysql.com/doc/refman/8.0/en/set-sql-log-bin.html: suggests that it only disables bin-logging of the current session

paulomach: from docs, looks precisely as what was needed

paulomach (approved):

PR #340, opened 2023-10-13
DPE-2662 single unit upgrade fixes

paulomach:
## Issue

1. single unit upgrade stuck at "other units upgrading first"
1. single unit won't reconnect to a cluster (as it is the cluster)
1. mysqlsh will fail on same version check

## Solution

1. upgrade library fix for single unit upgrade on vm at [PR#102](https://github.com/canonical/data-platform-libs/pull/102)
2. make usage of `reboot_from_complete_outage` for single unit deployment recovery
3. modify mysqlsh version verification command to accommodate same versions correctly


taurus-forever (dismissed):
We are loosing simplicity, but LGTM. Tnx!

taurus-forever (approved):

PR #349, opened 2023-11-09
Remove redundant step to clean CI disk space

shayancanonical:


paulomach (approved):

dragomirp (approved):

paulomach:
Don't remember the context here, but I'll close since it's stale. We can reopen if it's the case.

PR #354, opened 2023-11-15
Use reusable integration test workflow

carlcsaposs-canonical:


taurus-forever (approved):
Wow, this is way nicer now.

Q: why did we run unit test for Juju 2.9 and 3.1 earlier (no neither Juju nor libjuju used in unit tests)? I am confused here...

shayancanonical (approved):
:rocket:

PR #360, opened 2023-11-28
[DPE-2677] Avoid setting secret upon tls relation broken if using juju secrets

shayancanonical:
## Issue
We do not have access to a juju secret upon relation-broken event. However, the tls relation code in MySQL is setting tls related secrets to None. This is necessary for secrets stored in the databag (as they are reset to None), but causes the trace in the Jira ticket as the secret is deleted from the secrets backend

## Solution
Do not attempt to reset the tls secrets upon relation broken if using juju secrets

### DIFF
--- a/lib/charms/mysql/v0/tls.py
+++ b/lib/charms/mysql/v0/tls.py
@@ -22,6 +22,7 @@ It requires the TLS certificates library and the MySQL library.
 
 import base64
 import logging
+import ops
 import re
 import socket
 from typing import List, Optional, Tuple
@@ -52,7 +53,7 @@ LIBID = "eb73947deedd4380a3a90d527e0878eb"
 
 LIBAPI = 0
 
-LIBPATCH = 2
+LIBPATCH = 3
 
 SCOPE = "unit"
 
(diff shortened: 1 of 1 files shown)


paulomach:
> > We do not have access to a juju secret upon relation-broken event.
> 
> wait what? is this intended behavior by juju?

it make sense to do so for relation secrets, could not find documentation for that. @tmihoc here's a shameless plug for you!

> 
> also, should we be cleaning up the secret on relation broken?

Do you see any reason to keep it?

imo we should go fwd with the solution, and refactor if needed.

paulomach (approved):
Given the behavior, that's an appropriated solution

paulomach:
So, from the chat, since the charm owns the secret, it should not lose access? That does not seem to be what was observed

paulomach:
> checking behavior with juju team here: https://chat.charmhub.io/charmhub/pl/8u7d138khpdypmbxfw4uetxjbo
> 
> if the behavior is expected, I think this PR is probably a simpler workaround than checking if the unit is dying
> 
> @paulomach will the changes in this PR be okay if the relation with tls is removed and re-added? (the secret won't be cleaned up) if so, PR lgtm

It should, as the secret will always be overwritten (or written for 2.9) on relation re-added, with a new value.
The tls usage in the workload will be disabled when the relation is gone and re-enabled after receiving the new certificates.

PR #361, opened 2023-11-29
Remove ops from integration test dependencies

carlcsaposs-canonical:


taurus-forever (approved):
OK. Smells like has_secret should be a part of data-platform lib :-D

paulomach (approved):

PR #362, opened 2023-11-29
Update snap revision to 91 (MySQL v8.0.35)

shayancanonical:
## Issue
We release a new snap to the edge channel (revision 91) which has MySQL v8.0.35. However, our charm is using the snap revision 69 with MySQL v8.0.34

## Solution
Update snap revision to use the revision with MySQL v8.0.35


paulomach:
Additionally, for consistency, we should update `pyproject.toml`

```diff
- mysql-connector-python = "~8.0.33"
+ mysql-connector-python = "~8.0.35"
```

taurus-forever (approved):
Thank you!

PR #365, opened 2023-12-05
Set up logrotate when upgrading + account for logrotate's systemd timer

shayancanonical:
## Issue
1. We are not setting up logrotate's cron and config files when we upgrade from the previous stable (which does not have logrotate implemented)
2. Everyday, at midnight, the cron job running logrotate for us conflicts with the logrotate run by logrotate's systemd time

## Solution
1. Set up cron and logrotate config files when upgrade_granted
2. Change cron to avoid running at midnight

### DIFF
--- a/src/mysql_vm_helpers.py
+++ b/src/mysql_vm_helpers.py
@@ -257,7 +257,10 @@ class MySQL(MySQLBase):
         with open("/etc/logrotate.d/flush_mysql_logs", "w") as file:
             file.write(rendered)
 
-        cron = "* 1-23 * * * root logrotate -f /etc/logrotate.d/flush_mysql_logs\n1-59 0 * * * root logrotate -f /etc/logrotate.d/flush_mysql_logs\n"
+        cron = (
+            "* 1-23 * * * root logrotate -f /etc/logrotate.d/flush_mysql_logs\n"
+            "1-59 0 * * * root logrotate -f /etc/logrotate.d/flush_mysql_logs\n"
+        )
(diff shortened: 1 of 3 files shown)


src/mysql_vm_helpers.py
```diff
@@ -257,7 +257,7 @@ def setup_logrotate_and_cron(self) -> None:
         with open("/etc/logrotate.d/flush_mysql_logs", "w") as file:
             file.write(rendered)
 
-        cron = "* * * * * root logrotate -f /etc/logrotate.d/flush_mysql_logs\n"
+        cron = "* 1-23 * * * root logrotate -f /etc/logrotate.d/flush_mysql_logs\n1-59 0 * * * root logrotate -f /etc/logrotate.d/flush_mysql_logs\n"
```
paulomach, line 260: nit:
```suggestion
        cron = (
            "* 1-23 * * * root logrotate -f /etc/logrotate.d/flush_mysql_logs\n"
            "1-59 0 * * * root logrotate -f /etc/logrotate.d/flush_mysql_logs\n"
        )
```

paulomach (commented):
Tiny readability nit. This and the failing unit test and should be good to go.

src/mysql_vm_helpers.py
shayancanonical: addressed in [eea60e1](https://github.com/canonical/mysql-operator/pull/365/commits/eea60e1a82000e4218b7fd04f1340cd8b46abeca)

paulomach (approved):

PR #368, opened 2023-12-10
fix: ensure that poetry is available to tox

tonyandrewmeyer:
## Issue

If you don't have `poetry` installed locally (and on your PATH) then cloning the repo and running `tox` will fail (because tox tries to use poetry to set up the test requirements).

## Solution

Have tox install `poetry` as a dependency. I haven't pinned it at all - I'm not sure if you would prefer to do that or not (but currently you're using whatever `poetry` people have on their system, or is in the CI runner, so that's also unpinned in a way). I can do that if you pick a version, or feel free to adjust it yourself of course.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -24,6 +24,8 @@ allowlist_externals =
     {[testenv]allowlist_externals}
     charmcraft
     mv
+deps:
+    poetry
 commands_pre =
     poetry export --only main,charm-libs --output requirements.txt
 commands =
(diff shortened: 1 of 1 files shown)


taurus-forever (approved):
Thank you for the PR! LGTM. Requested review by MySQL squad.

paulomach (approved):
Thanks @tonyandrewmeyer

PR #383, opened 2024-01-22
fix: when removing a secret, remove it from the cache, too

tonyandrewmeyer:
## Issue

When a secret is completely emptied of content, the lib removes all revisions from Juju, but leaves the cached object in place. This means that code can still access it, when in practice it has actually gone.

## Solution

Remove the secret from the `SecretCache` when it's removed from Juju with `remove_all_revisions()`. This does access a private attribute, but the comment just above this says that this code is temporary until the removal is in the lib itself, where the private attribute access would be fine.

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -651,6 +651,7 @@ class MySQLCharmBase(CharmBase, ABC):
             secret.set_content(content)
         else:
             secret.meta.remove_all_revisions()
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer:
I'm not sure on the protocols here, sorry - am I meant to do an update of the libraries even though that's not required here? Happy to do so if that's the case.

tonyandrewmeyer:
It seems like this code has moved into data-platform-libs so this is now out-of-date.

PR #400, opened 2024-02-21
Add first performance options

phvalguima:
Some of the early conclusions I've found. These options will not use memory maps for files but rather allow "direct" access. The advantage: mysql has its own engines to cache data and does not need the kernel to manage file caches itself.


taurus-forever (approved):
IMHO, good default as we are deploying DB.
@delgod to decide on the architecture level... should we sit on DB default or tune them as we think we should.
Tnx Pedro!

paulomach (approved):
lgtm, just remember to bump the libpatch value and fix unit tests

paulomach (approved):

taurus-forever:
Hi,

JFYI, I have made 1 hour tests on `localhost` for default and proposed parameters (twice).
Env: Laptop=>Multipass QEMU=>LXD.
For the small numbers (like 6K qpt) the difference is not really visible:
Screenshot from 2024-03-11 09-40-13

I still believe we should go with new defaults based on the previous test results done by Pedro.

P.S. Ignore significant spikes/drops on the localhost test, it looks like some laptop background tasks. No issues notices on charm/qemu itself.

taurus-forever (approved):

sinclert-canonical:
Repository moved to a new home: https://github.com/canonical/mysql-operators.

PR #403, opened 2024-02-28
DPE-3688 fix KeyError when no ca-chain

### LINKED ISSUE #402: Backup test failing with backup-id not found
Backup fail to execute due missing "tls-cha-chain" key in s3_helper s3_parameters

paulomach:
## Issue

backup fail due to dict key error

## Solution

- avoid KeyError when no ca chain
- Increase coverage

Fixes #402

### DIFF
--- a/lib/charms/mysql/v0/s3_helpers.py
+++ b/lib/charms/mysql/v0/s3_helpers.py
@@ -50,9 +50,9 @@ def upload_content_to_s3(content: str, content_path: str, s3_parameters: Dict) -
 
     Returns: a boolean indicating success.
     """
+    ca_file = None
     try:
         logger.info(f"Uploading content to bucket={s3_parameters['bucket']}, path={content_path}")
-        ca_file = tempfile.NamedTemporaryFile()
         session = boto3.session.Session(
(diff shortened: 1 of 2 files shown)


lib/charms/mysql/v0/s3_helpers.py
```diff
@@ -86,7 +86,8 @@ def upload_content_to_s3(content: str, content_path: str, s3_parameters: Dict) -
         )
         return False
     finally:
-        ca_file.close()
+        if ca_file:
+            ca_file.close()
```
taurus-forever, line 90: Why don't we close right after ca_file.flush() ? Do you also want to close `temp_file`?
Anyway, this will work.

taurus-forever (approved):
Tnx for tests!

lib/charms/mysql/v0/s3_helpers.py
paulomach: The `NamedTemporaryFile` will be deleted once closed, hence closing it by the end.
The `temp_file` is used within a context manager (`with`), which closes it when exits the context

PR #416, opened 2024-03-15
DPE-3830 - Paxos single leader

paulomach:
Use paxos consensus algorithm in single leader mode as default.
Per documentation:

> Operating with a single consensus leader improves performance and resilience in single-primary mode, particularly when some of the group’s secondary members are currently unreachable

Reference at [link](https://dev.mysql.com/doc/refman/8.0/en/group-replication-system-variables.html#sysvar_group_replication_paxos_single_leader)


lib/charms/mysql/v0/mysql.py
```diff
@@ -1023,6 +1023,7 @@ def render_mysqld_configuration(  # noqa: C901
             "log_error": f"{snap_common}/var/log/mysql/error.log",
             "general_log": "OFF",
             "general_log_file": f"{snap_common}/var/log/mysql/general.log",
+            "loose-group_replication_paxos_single_leader": "ON",
```
sinclert-canonical, line 1026: Minor: could we group all `loose-` prefixed config options together? That way it would be simpler to quickly identify which subset of config options are dropped when passing them to `mysqld`.

sinclert-canonical (approved):

lib/charms/mysql/v0/mysql.py
paulomach: Oh, saw after merge. But yes, we can do it anytime

PR #417, opened 2024-03-18
[DPE-3702] Add snap aliases for MySQL server / Router

lucasgameiroborges:
## Issue

Missing snap aliases (ie using <snap> instead of charmed-msql.<snap>) for better tool usability inside VM.

## Solution

Heavily inspired by [ticket DPE-2090](https://github.com/canonical/postgresql-operator/pull/170).

## Testing

Since it is my first PR, here's how I went about testing my changes. Any feedback is appreciated!

```
# building local charm
tox run -e build-wrapper && mv requirements-last-build.txt requirements.txt && charmcraft pack

# deploy local charm
juju deploy ./mysql_ubuntu-22.04-amd64.charm 

# ssh into VM
juju ssh mysql/0 bash

# seeing snap aliases
snap aliases
```

With output:
```
Command                    Alias        Notes
charmed-mysql.mysql        mysql        manual
charmed-mysql.mysqlrouter  mysqlrouter  manual
charmed-mysql.mysqlsh      mysqlsh      manual
charmed-mysql.xbcloud      xbcloud      manual
charmed-mysql.xbstream     xbstream     manual
charmed-mysql.xtrabackup   xtrabackup   manual
lxd.lxc                    lxc          -
```

### DIFF
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -409,9 +409,9 @@ class Snap(object):
         """Remove the refresh hold of a snap."""
         self._snap("refresh", ["--unhold"])
 
-
     def alias(self, application: str, alias: Optional[str] = None) -> None:
         """Create an alias for a given application.
+        
         Args:
             application: application to get an alias.
             alias: (optional) name of the alias; if not provided, the application name is used.
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -83,7 +83,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 12
+LIBPATCH = 13
 
 
 # Regex to locate 7-bit C1 ANSI sequences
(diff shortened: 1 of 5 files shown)


lib/charms/operator_libs_linux/v1/snap.py
dragomirp, line 1: This is an external library coming from another [charm](https://charmhub.io/operator-libs-linux) ([repo](https://github.com/canonical/operator-libs-linux)), it is not something we should be editing directly. The aliasing functionality should be available in the v2 of the lib. The way to update it is:
```
# Charmcraft needs requirements.txt to exist so generate it
> tox -e build-wrapper
> mv requirements-last-build.txt requirements.txt
# Fetch the lib
> charmcraft fetch-lib charms.operator_libs_linux.v2.snap
```
Then you can remove the old lib (v1) and update imports. IIRC there were no interface changes from v1 to v2.

tests/unit/test_charm.py
dragomirp, line 1: We should assert that the expected alias calls were made.

dragomirp:
> I see the `check libraries` test failing in every commit, including in other PRs. Is this expected?

I think it's expected, since there are other outdated charm libraries.

tests/unit/test_charm.py
lucasgameiroborges: missed this before, it should be good now, thanks for catching

taurus-forever (approved):
Nice, thank you for the first contribution !!!

shayancanonical (approved):
we are creating snap aliases for all commands in [the snap's apps](https://github.com/canonical/charmed-mysql-snap/blob/8.0/edge/snap/snapcraft.yaml#L75), but not the daemons. LGTM!

lucasgameiroborges:
I think I'll go ahead and merge it, then. Thanks everyone!

PR #428, opened 2024-03-28
Add extended attribute to get-cluster-status

arif-ali:
This changes from the default boolean that was being used internally to what mysql expects. The values between 0 and 3 are valid.

## Issue

Would like to get extra cluster status attributes to debug issues with mysql replication.

## Solution

The solution allows us to do this via actions, rather than logging in, and getting the data manually


actions.yaml
```diff
@@ -3,6 +3,11 @@
 
 get-cluster-status:
   description: Get cluster status information
+  params:
+    extended:
+      type: integer
+      description: The extended attriute for cluster-status, the default value 0.
```
paulomach, line 9:
```suggestion
      description: The extended attribute for cluster-status, the default value 0.
```

actions.yaml
```diff
@@ -3,6 +3,11 @@
 
 get-cluster-status:
   description: Get cluster status information
+  params:
+    extended:
+      type: integer
```
paulomach, line 8:
```suggestion
      type: integer
      default: 0
```

lib/charms/mysql/v0/mysql.py
```diff
@@ -494,7 +494,9 @@ def _on_set_password(self, event: ActionEvent) -> None:
 
     def _get_cluster_status(self, event: ActionEvent) -> None:
         """Action used  to retrieve the cluster status."""
-        if status := self._mysql.get_cluster_status():
+        extended = event.params.get("extended") or 0
```
paulomach, line 497:
```suggestion
        extended = event.params.get("extended", 0)
```

lib/charms/mysql/v0/mysql.py
```diff
@@ -494,7 +494,9 @@ def _on_set_password(self, event: ActionEvent) -> None:
...
+
```
paulomach, line 498: value need validation, e.g.
```suggestion
if not 0 <= extended < 4:
  event.fail("Extended parameter outside valid range")
  return

```

paulomach (changes requested):
Hi @arif-ali, 

thank you for your contribution.
Please, check comments

taurus-forever (approved):
Thank you for the contribution here! 
@paulomach will we copy to K8s ourself or should we ask @arif-ali ?

we do it

lib/charms/mysql/v0/mysql.py
```diff
@@ -494,7 +494,13 @@ def _on_set_password(self, event: ActionEvent) -> None:
 
     def _get_cluster_status(self, event: ActionEvent) -> None:
         """Action used  to retrieve the cluster status."""
-        if status := self._mysql.get_cluster_status():
+        extended = event.params.get("extended", 0)
```
shayancanonical, line 497: the libpatch on this file should be incremented

shayancanonical (changes requested):

lib/charms/mysql/v0/mysql.py
paulomach: @arif-ali , I thinks that should be the only pending change.

arif-ali: That should be done on the latest push

paulomach (approved):
lgtm

sinclert-canonical:
👋🏻 Are these changes still desired? They got approved one year ago, despite Shayan's request to bump the lib version.

sinclert-canonical:
I can probably help with that, this PR seems like a low hanging fruit everyone was onboard with. Although, if we are going to do it right, the same changes should be proposed both in K8s and VM substrates.

PR #429, opened 2024-03-28
DPE-3706 Fix max_connections calculation and expose experimental max_connections field

### LINKED ISSUE #407: `innodb_buffer_pool_size` and `max_connections` are trying to use ~175% of the physical memory?
(there is question mark in the title since I'm not 100% sure yet before running a stress test)

## Steps to reproduce

1. juju deploy mysql --channel 8.0/edge

## Expected behavior

mysqld should allocate memory for `innodb_buffer_pool_size` and `max_connections` within the amount of the physical memory available.

## Actual behavior

The charm writes the following config with 32GB memory system.

[/var/snap/charmed-mysql/current/etc/mysql/mysql.conf.d/z-custom-mysqld.cnf]
```
[mysqld]
bind-address = 0.0.0.0
mysqlx-bind-address = 0.0.0.0
report_host = juju-584fb0-0.lxd
max_connections = 2570
innodb_buffer_pool_size = 23219666944
log_error_services = log_filter_internal;log_sink_internal
log_error = /var/snap/charmed-mysql/common/var/log/mysql/error.log
general_log = ON
general_log_file = /var/snap/charmed-mysql/common/var/log/mysql/general.log
slow_query_log_file = /var/snap/charmed-mysql/common/var/log/mysql/slowquery.log
innodb_buffer_pool_chunk_size = 2902458368
```

- available_memory: 32,341,442,560 (~32GB)
- innodb_buffer_pool_size = 23,219,666,944 (~23GB)
- innodb_buffer_pool_chunk_size = 2,902,458,368 (~3GB)
- max_connections = 2,570 -> 2,570 * 12MiB (12,582,912) = 32,338,083,840 (~32GB)

So it looks like innodb_buffer_* tries to use ~23GB of memory at maximum and max_connections tries to use ~32GB of memory at maximum on the same 32GB memory.

https://github.com/canonical/mysql-operator/blob/8bf9fb36ce23b06f5b0838868a2ebe54486d912f/lib/charms/mysql/v0/mysql.py#L717-L727

mysql_major_parameters

## Versions

Operating system: Ubuntu 22.04 LTS

Juju CLI: 3.4.0-genericlinux-amd64

Juju agent: 3.4.0

Charm revision: mysql  8.0/edge  210

LXD: 5.20

## Log output

Juju debug log: 

This is from an extra debug logging with by hand modification to add some traces.
```
2024-03-05 14:13:28 WARNING unit.mysql/0.juju-log server.go:325 available_memory=32341442560
2024-03-05 14:13:28 WARNING unit.mysql/0.juju-log server.go:325 memory_limit=0
2024-03-05 14:13:28 WARNING unit.mysql/0.juju-log server.go:325 available_memory=32341442560
2024-03-05 14:13:28 WARNING unit.mysql/0.juju-log server.go:325 innodb_buffer_pool_size=23219666944
2024-03-05 14:13:28 WARNING unit.mysql/0.juju-log server.go:325 innodb_buffer_pool_chunk_size=2902458368
2024-03-05 14:13:28 WARNING unit.mysql/0.juju-log server.go:325 group_replication_message_cache_size=None
2024-03-05 14:13:28 WARNING unit.mysql/0.juju-log server.go:325 max_connections=2570
```

paulomach:
## Issue

Max connections calculations is not considering already allocated memory for `innodb_buffer_pool_size`, resulting in overcommit (and OOM kill risk)

## Solution

* Calculate `max_connections` based on the available memory after allocating `innodb_buffer_pool_size`.
* Expose experimental `experimental-max-connections` configuration allowing independent tuning (as agreement with PM in Madrid)

Fixes #407


lib/charms/mysql/v0/mysql.py
```diff
@@ -739,7 +739,12 @@ def render_mysqld_configuration(
                 innodb_buffer_pool_chunk_size,
                 group_replication_message_cache_size,
             ) = self.get_innodb_buffer_pool_parameters(available_memory)
-            max_connections = max(self.get_max_connections(available_memory), MIN_MAX_CONNECTIONS)
+
+            # constrain max_connections based on the available memory
+            # after innodb_buffer_pool_size calculation
+            remaining_memory = available_memory - innodb_buffer_pool_size
+            max_connections = max(self.get_max_connections(remaining_memory), MIN_MAX_CONNECTIONS)
```
shayancanonical, line 746: i see the logic behind this, but are we sure we want to limit `max_connections` based on the remaining memory after allocation about 75% to 50% to `innodb_buffer_pool_size`? would it bite us if the max_connections is severly limited compared to client expectations?

paulomach: I agree, and we probably can over commit here. But I'll defer that to @taurus-forever and @delgod

nobuto-m: Essentially this is (at least) a 3-parameter problem to me. The proposed logic is better than the current one since it can avoid overcommitting memory to 175%. However, keeping limiting user-configurable parameters to 1 doesn't solve the fundamental challenges.

mysql_major_parameters

At this moment, we don't allow configuration other than the parameter "A" in the diagram. I assume that the `mysql.py` logic is shared between the machine charm and the k8s charms so the parameter "A" is determined either by the total amount of physical memory, memory_limit charm config, or pod resource limit. But no other parameters about the memory allocation can be supplied by an user.

When an user needs to bump the max_connections we will have to bump the memory requirement 4 times unnecessarily since there is no way to tell the charm to change the ratio of the memory allocation (75% innodb pool size + 25% max_connections).

For example with MySQL use cases as the backend of OpenStack deployment, we know that it requires a few thousand max connections to cover the distributed (and micro service-ish) control plane services and the multiple workers of each micro service, but 8GB of innodb buffer pool size is more than enough.

Let's say to have max_connections=4,000, 64 GB of the instance is more than enough (4,000 connections * 12MB + 8GB innodb buffer pool size = 56GB). However, with the fixed ratio, we need 200GB machine (4,000 connections * 12MB * (75+25)% / 25% = 192GB).

I cannot think of a better way than moving away from a single parameter.

taurus-forever: TL;DR: @nobuto-m is correct.

My proposal was a [custom profile tuning](https://docs.google.com/document/d/1pymnut3bB7zhV6hggdNiHAU2VkONZYsoRsqELIzVBkQ/edit) which could be extended to `custom-mem-42%` + `custom-connections-100500` to fine tune the corner case scenarios.

Anyway, by default all pre-allocated memory like `innodb_buffer_pool_size` (and `group_replication_message_cache_size`?) should be excluded from available memory "reserved" for client connections. So this PR still necessary to avoid OOM/crashes.

CC: @7annaba3l , @delgod

taurus-forever (approved):

lib/charms/mysql/v0/mysql.py
delgod: @nobuto-m is correct.

but before decidion made on independent config options, we need to fix this OOM

nobuto-m:
Hi all, after discussing about this here and internally, I think I should request this patch to be on hold for the time being.

The fundamental challenge is not immediately solved at this point, and the proposed patch is good as is to prevent the OOM from happening. However, it will limit the available max_connections to 25% from the current state and that may make it harder to run workload with a machine with a fair amount of memory. Until we fix the original challenge, I think it might be better to leave it as is since 1. OOM doesn't happen immediately after the service start and 2. MySQL dynamically allocates memory for actually data in innodb buffer pool size and for incoming connections. So we can keep this optimistic way of memory allocation for the time being.

In the meantime, I appreciate @paulomach for taking the initiative on this matter so far.

delgod (changes requested):
please don't merge without my explicit approval

lib/charms/mysql/v0/mysql.py
```diff
@@ -871,12 +874,33 @@ def render_mysqld_configuration(
...
-            max_connections = max(self.get_max_connections(available_memory), MIN_MAX_CONNECTIONS)
+
+            # constrain max_connections based on the available memory
+            # after innodb_buffer_pool_size calculation
+            available_memory -= innodb_buffer_pool_size + (
+                group_replication_message_cache_size or 0
+            )
+            if not experimental_max_connections:
+                max_connections = max(
+                    self.get_max_connections(available_memory), MIN_MAX_CONNECTIONS
```
shayancanonical, line 901: nit (feel free to ignore): `MIN_MAX_CONNECTIONS` is not a minimum, rather a maximum where it is used (if max_safe_connections is < 100, we will bump max_connections up to 100). may be a better name would be `SUGGESTED_MAX_CONNECTIONS` or `MAX_CONNECTIONS_FLOOR` or something like that? 

also out of curiosity, what is the reasoning behind at least 100 max_connections?

shayancanonical (approved):

lib/charms/mysql/v0/mysql.py
paulomach: I'm happy to make it clearer. Right now, the value is guesswork, being smaller than default (151)

taurus-forever:
> please don't merge without my explicit approval

@delgod is it still valid blocker?

config.yaml
```diff
@@ -30,3 +30,11 @@ options:
   mysql-interface-database:
     description: "The database name for the legacy 'mysql' relation"
     type: "string"
+# Experimental features
+  experimental-max-connections:
```
nobuto-m, line 34: I'd like to challenge the config option name. I totally understand that it's experimental and it could be removed at any point so clarifying it in the config description is fine.

However, adding the prefix `experimental-` to the config option itself is a little bit awkward. Because the max-connections in MySQL itself is a quite known config and it's not experimental or anything like that. The idea behind the config naming is to indicate that "this option could disappear anytime once the charm gets capable to make the best decision" so I prefer a different name something like `max-connections-override` or something like that to indicate "while charm will try to make the best decision to configure max connections automatically, you can still override the value explicitly".

lib/charms/mysql/v0/mysql.py
paulomach: 013a6819

delgod (approved):

config.yaml
delgod: @nobuto-m, Thank you for your feedback!
We have decided to use the `experimental-` prefix on a Product level to make it consistent across multiple solutions. This decision affects not only `max-connections` but also a few other things where `override` doesn't work.
Please provide product feedback or discuss directly with decision-makers.

delgod (approved):

PR #430, opened 2024-04-03
[DPE-3824] Upgrade mysqld-exporter to v0.15

shayancanonical:
## Issue
We would like to test a new build of the charmed-mysql-snap which contains an upgraded version of mysqld-exporter

## Solution
Update snap revision to reflect the new build

## TODO
Update charm revision to `8.0/edge` revision after https://github.com/canonical/charmed-mysql-snap/pull/45 is merged in


shayancanonical:
Closing as PR is irrelevant right now - mysqld-exporter has the following issues https://github.com/prometheus/mysqld_exporter/issues/754 that we are running into

1. mysqld-exporter 0.15 does not correctly parse unix sockets as `--mysqld.address` (it expects host:port)
2. mysqld-exporter 0.15 expects my.cnf style file with args even though it should be optional if the args are passed in through the command line

PR #451, opened 2024-05-03
DPE-4273 avoid restart being triggered after upgrade

paulomach:
## Issue

* on upgrade, config file was not being update/override in the process
* after updgrade, config change was triggering restart of the service
* restart hook failed to consider all units

This was causing the failure of scaling up after a successful upgrade.

## Solution

* addressed issues above


taurus-forever (approved):
LGTM, also tested locally, cannot reproduce DPE-4273 anymore. Tnx!

PR #462, opened 2024-06-13
[DPE-4118] Add support for rescanning cluster for unit rejoin after node drain

shayancanonical:
Part of https://github.com/canonical/mysql-k8s-operator/pull/433


lib/charms/mysql/v0/mysql.py
```diff
@@ -1599,6 +1599,24 @@ def initialize_juju_units_operations_table(self) -> None:
             )
             raise MySQLInitializeJujuOperationsTableError(e.message)
 
+    def rescan_cluster_if_instance_already_in_cluster(
```
paulomach, line 1602: I would prefer to modify the existing method `in_instance_in_cluster` to have the `from_instance` parameter, and test for member states there, so we can ditch this new method.

Then call both methods (is_instance_in_cluster and rescan) within the charm code, so the flow is clear there

paulomach (changes requested):
Left couple comments. Also include unit testing

paulomach (approved):

PR #466, opened 2024-06-21
Upgrade from TLS charm lib v2 to v3

shayancanonical:
## Issue
We are considering using the `CertHandler` to integrate with a certificates operator to support exporting traces to tempo with HTTPS. However, `CertHandler` uses `tls_certificates` charm lib v3, while our charm uses charm lib v2.

## Solution
Upgrade the TLS charm lib from v2 to v3 in our charm


shayancanonical:
Blocker: we cannot upgrade to tls certificates charm lib v3 because it requires secrets: https://github.com/canonical/mysql-operator/actions/runs/9617004380/job/26528063502?pr=466#step:21:232

PR #471, opened 2024-06-25
DPE-4699 backup logging and other fixes

paulomach:
## Issue

* no realtime logs for long running sub commands
* lack of status for users
* restore timeouts if client session has pending transaction
* post_restore was not creating cluster-set
* flush logs are attempted when mysqld is not running

## Solution

* stream backup/store related commands to juju log
* improved unit status setting
* kill client sessions on pre-restore
* add cluster-set creation to post-restore
* only flush logs if mysqld is running
* typing fixes here and there


src/mysql_vm_helpers.py
```diff
@@ -527,31 +529,54 @@ def _execute_commands(
...
+            stderr=subprocess.PIPE,
+        )
+
+        if stream_output == "stderr":
+            while process.stderr and (line := process.stderr.readline()):
+                logger.debug(line.strip())
+                stderr += line
+        elif stream_output == "stdout":
+            while process.stdout and (line := process.stdout.readline()):
+                logger.debug(line.strip())
```
shayancanonical, line 561: very nice!

shayancanonical (approved):
Wonderful stuff!

lib/charms/mysql/v0/backups.py
```diff
@@ -533,22 +543,29 @@ def _on_restore(self, event: ActionEvent) -> None:
...
+    def _pre_restore(self) -> Tuple[bool, str]:
         """Perform operations that need to be done before performing a restore.
 
         Returns: tuple of (success, error_message)
         """
+        if not self.charm._mysql.is_mysqld_running():
+            return True, ""
+
         try:
+            logger.info("Stopping mysqld before restoring the backup")
```
taurus-forever, line 558: Should we remove `endpoint` from relation -> informing router/clients DB is not reachable?

taurus-forever (approved):
What would be amazing: merge and play it edge. Excellent work and nice UX!

P.S. I was expecting a bit smarter backup tests separation (reusing code, as you have copied 377 lines into the second file changing there 4 constants :-)... microceph has 9 lines difference (from 377)... supporting those copy&paste will produce human mistakes. 

Up to you.

PR #478, opened 2024-07-03
DPE-4699 Set instance offline mode on restore

paulomach:
## Issue

Clients will try reconnect to server on restore, which can prevent service stop

## Solution

Ensure clients can connect by setting server offline mode


tests/integration/high_availability/test_async_replication.py
```diff
@@ -179,6 +179,7 @@ async def test_deploy_router_and_app(first_model: Model) -> None:
         application_name=MYSQL_ROUTER_APP_NAME,
         series="jammy",
         channel="dpe/edge",
+        revision=202,  # TODO: Remove once router is published
```
shayancanonical, line 182: is this still necessary?

shayancanonical (approved):

PR #482, opened 2024-07-11
DPE-4850 cluster metadata and instance state checks

paulomach:
## Issue

See k8s [issue](https://github.com/canonical/mysql-k8s-operator/issues/457) and [Pr](https://github.com/canonical/mysql-k8s-operator/pull/458)

## Solution

* implemented changes for fixes on k8s
* refactored lib/charm to be less reliant on databag keys
* commentary cleanup for size reduction (<128k)

PR #483, opened 2024-07-11
Attempt to reduce charm lib size so it can be published

shayancanonical:
## Issue
We cannot publish our mysql charm lib because it is too large in size (https://github.com/canonical/mysql-operator/actions/runs/9884439921/job/27315469186#step:4:82)

## Solution
Remove docstring args, kwargs, raises and return info


paulomach (approved):

paulomach:
replaced by #482

PR #484, opened 2024-07-15
Remove unused key/values from the databag

shayancanonical:
## Issue
We refactored how we calculate `cluster_initialized` and `unit_initialized` in #482. This uses the metadata in the database instead of relying on `units-added-to-cluster` and `unit-intialized` values in the databag. However, there are still places in the code where we set these keys

## Solution
Remove statements that set these keys

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -89,6 +89,7 @@ jobs:
       cloud: lxd
       juju-agent-version: ${{ matrix.juju.agent }}
       libjuju-version-constraint: ${{ matrix.juju.libjuju }}
+      lxd-snap-channel: 5.21/stable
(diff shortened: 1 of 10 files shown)


paulomach (commented):
I've intended to leave `units-added-to-cluster` to trigger `peer-relation-changed` and speed up deployment, since it handle other units being added. But thinking it through, it may not be necessary. 
Did you notice any slowdown?

paulomach:
> Since we are setting `member-state` and `member-role`, peer-relation-changed gets triggered anyway and there is no delay in joining the cluster

so kill them with :fire:

paulomach (approved):

taurus-forever (approved):
Great to keep the house clean!

PR #487, opened 2024-07-19
[DPE-4118] Port mysql lib changes from k8s + increment tls lib for local changes

shayancanonical:
## Issue
1. We need to break out a method for executing remove_instance commands (changes already in [k8s PR](https://github.com/canonical/mysql-k8s-operator/pull/433)
2. We need to increment the tls lib as there were changes in [PR 482](https://github.com/canonical/mysql-operator/pull/482), but libpatch wasnt updated

## Solution
1. Port over changes from the k8s PR
2. Increment tls lib's libpatch

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -128,7 +128,7 @@ LIBID = "8c1428f06b1b4ec8bf98b7d980a38a8c"
 # Increment this major API version when introducing breaking changes
 LIBAPI = 0
 
-LIBPATCH = 63
+LIBPATCH = 64
 
 UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
 UNIT_ADD_LOCKNAME = "unit-add"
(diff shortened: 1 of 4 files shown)


paulomach (approved):
lgtm, just add a unit test for the new method please

lib/charms/mysql/v0/mysql.py
```diff
@@ -1778,6 +1778,24 @@ def _get_host_ip(host: str) -> str:
...
 
+    def execute_remove_instance(self, connect_instance: Optional[str] = None) -> None:
+        """Execute the remove_instance() script with mysqlsh.
+
+        Args:
+            connect_instance: (optional) The instance from where to run the remove_instance()
+        """
+        remove_instance_options = {
+            "password": self.cluster_admin_password,
+            "force": "true",
```
taurus-forever, line 1789: Q: do we need to force it by default?

taurus-forever (approved):
LGTM, just one question.

BTW, I would merge LIBPATH ASAP if this PR stuck. Otherwise it can be polluted by https://github.com/canonical/mysql-operator/pull/489/files#diff-414e9b4c59325e09514ab3b986f01bd805dcecd3b568d5ff45a46e357416531d

lib/charms/mysql/v0/mysql.py
shayancanonical: both of our use cases need force, but added a kwarg for `force=False` (tested against https://github.com/canonical/mysql-k8s-operator/actions/runs/10048296233)

PR #493, opened 2024-07-25
[DPE-2477] Add a minimal set of alert rules from the Awesome Prometheus Alert Rules blog

shayancanonical:
## Issue
We would like to import a minimal set of alert rules from the [Awesome Prometheus Alert Rules blog](https://samber.github.io/awesome-prometheus-alerts/rules#mysql)

## Solution
Import rules (noting slight modifications). Do not include any replication rules as they are not Group Replication related

image

### DIFF
--- a/src/alert_rules/prometheus/metrics_alert_rules.yaml
+++ b/src/alert_rules/prometheus/metrics_alert_rules.yaml
@@ -14,7 +14,7 @@ groups:
 
       # 2.1.2
       # customized: 80% -> 90%
-      - alert: MySQLTooManyConnections(>80%)
+      - alert: MySQLTooManyConnections(>90%)
         expr: max_over_time(mysql_global_status_threads_connected[1m]) / mysql_global_variables_max_connections * 100 > 90
         for: 2m
         labels:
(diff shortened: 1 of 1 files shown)


src/alert_rules/prometheus/metrics_alert_rules.yaml
```diff
@@ -11,3 +12,65 @@ groups:
...
 
+      # 2.1.2
+      # customized: 80% -> 90%
+      - alert: MySQLTooManyConnections(>80%)
+        expr: max_over_time(mysql_global_status_threads_connected[1m]) / mysql_global_variables_max_connections * 100 > 90
+        for: 2m
+        labels:
+          severity: warning
+        annotations:
+          summary: MySQL too many connections (> 90%) (instance {{ $labels.instance }})
```
taurus-forever, line 23: It is 80% in lines 17-18 and 90% here. Typos?

taurus-forever (commented):
LGTM as a step 1, one typo inside otherwise LGTM. Thank you!

src/alert_rules/prometheus/metrics_alert_rules.yaml
shayancanonical: oops, sorry, yes typo. corrected in 029f4b826677a277e3f935c317413ecaacf108e4

paulomach (approved):
Ok for starting. We need to check if not having GR alerts/metrics is a show stopper though

PR #496, opened 2024-08-02
Add charmcraft part to prime files

carlcsaposs-canonical:
Support for `prime` in `charm` part removed in charmcraft 3


taurus-forever (approved):
It should be deprecated only and should still work... but LGTM to be the future compatible.

paulomach (approved):

PR #503, opened 2024-08-07
DPE-4247 Change binlog retention period

paulomach:
## Issue

Default binlog retention period (30 days) not aligned with other logs retention period (7 days).
Change requested by support team due huge binlogs on Kubeflow deployments.

## Solution

Set binlog retention period to 7 days.


taurus-forever (approved):
@delgod , @7annaba3l please approve!

LGTM, but we might need to improve "re-joining the cluster". When the unit is down longer the retention period, starting unit will not re-join the cluster without the full DB copy (as no binlogs to read from).

Do we need a config option here? Telco normally are not allowed to delete binlogs, they are archived and moved to external storage for the rare cases investigations.

7annaba3l:
This is approved on my end.
@taurus-forever : Indeed, we should archive bin logs  to an ObjectStore after the  retention period. 
Hopefully, this will be part of our backup manager implementation.
For now, we can provide a config parameter to provide binlog retention in days (with a minimum of 1).

shayancanonical (approved):

PR #505, opened 2024-08-09
Install/uninstall plugin in single session

paulomach:
## Issue

Toggling `super_read_only` on a session previous to plugin install can fail due to `super_read_only` toggled by another database process. Toggling is necessary for install plugin on secondaries. 

## Solution

Modify install/uninstall methods to run in a single session.


taurus-forever (approved):
LGTM, tnx!

shayancanonical (approved):

PR #508, opened 2024-08-15
Stabilize integration tests; avoid using equality for comparison of units and compare unit names instead

shayancanonical:
## Issue
Network cut test: https://github.com/canonical/mysql-operator/actions/runs/10397182320/job/28792633963
Replicate data on restart test: https://github.com/canonical/mysql-operator/actions/runs/10397182320/job/28792634154#step:28:425

Both of the above tests fail on retrieving a cluster primary when the old primary is stopped or cut from the network. The reason for this, I believe, is the usage of `==` to compare units. Despite the `unit` and `unit_excluded` is the same, the unit is not excluded (the condition does not hold true). This is why the tests above fail to get primary from the "old primary unit" which does not have a running mysqld.

## Solution
Compare unit names instead

### DIFF
--- a/lib/charms/tempo_k8s/v1/charm_tracing.py
+++ b/lib/charms/tempo_k8s/v1/charm_tracing.py
@@ -224,7 +224,6 @@ def _remove_stale_otel_sdk_packages():
 
 _remove_stale_otel_sdk_packages()
 
-
 import functools
 import inspect
 import logging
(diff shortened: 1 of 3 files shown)


paulomach (approved):
good catch

taurus-forever (approved):
Excellent catch. If you assumplion is true, please share this finding with entire Data team!

PR #516, opened 2024-09-02
DPE-5178 Adopt admin-address throught out

paulomach:
## Issue

As part of cycle, we ought to utilize admin-address for charm usage.
The idea is to ensure that the charm can connect to mysqld
instances (local or remote unit) even when `max_connections` is saturated
by related application(s), since connections to admin-address are do not count against `max_connections`.

## Solution

* library code to render `admin_address` in configuration file.
* use `admin_address:admin_port` for users with `ADMIN_CONNECTION` privilege.
* `s/clusteradmin/serverconfig` due cluster admin not having (and not needing to have, since replication channels don't count against `max_connection`) `ADMIN_CONNECTION` privilege.
* add integration test to saturate client connections
* use new feature on mysql-test-app [PR#37](https://github.com/canonical/mysql-test-app/pull/37) to retrieve connection data, used in integration test


tests/integration/connector.py
```diff
@@ -37,3 +39,30 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+        conn = mysql.connector.connect(
+            host=host,
+            user=username,
+            password=password,
+            database=database,
+            use_pure=True,
+        )
+        if conn.is_connected():
+            connections.append(conn)
+        sleep(0.5)
```
paulomach, line 67: sleep added to slow down tests so I could validate connections in database. Left here for the same reason, since the test is fairly quick

tests/integration/test_saturate_max_connections.py
```diff
@@ -0,0 +1,92 @@
...
+    config = {"auto_start_writes": False, "sleep_interval": "500"}
+    logger.info("Deploying test app")
+    await ops_test.model.deploy(
+        "mysql-test-app",
+        application_name=TEST_APP_NAME,
+        num_units=1,
+        base="ubuntu@22.04",
+        config=config,
+        channel="latest/edge",
+        revision=55,  # test branch
```
paulomach, line 49: remove before merge and after test-app merge

tests/integration/test_saturate_max_connections.py
```diff
@@ -0,0 +1,92 @@
...
+    assert len(connections) == CONNECTIONS, "Not all connections were established"
+
+    logger.info("Ensure no more client connections are possible")
+
+    with pytest.raises(OperationalError):
+        # exception raised when too many connections are attempted
+        create_db_connections(1, **credentials)
+
+    logger.info("Get cluster status while connections are saturated")
+    _ = await run_action(mysql_unit, "get-cluster-status")
```
shayancanonical, line 92: should we also add a check that the admin port is unaffected? or at least admin connections are active?

shayancanonical (approved):
Looks great, left a comment about asserting that the admin connection is possible even when max connections are saturated

tests/integration/test_saturate_max_connections.py
paulomach: Getting the cluster status rely on the admin connection. So when client connections are saturated, the action passing does just that. If the action fails, `run_action` will raise an assertion error

tests/integration/connector.py
taurus-forever: It smells to me, it is a workaround for https://warthogs.atlassian.net/browse/DPE-5340 (we should sent peer details when we ready to accept traffic only).

taurus-forever (approved):
Thank you! LGTM.
Q: do we still have usecases for clusteradmin OR we need to clean it one day in the future from Juju Secrets and Documentation?

tests/integration/connector.py
paulomach: Not a workaround for that issue. Just to be able to observe the test while developing.
The error on the nightly tests seems a different issue to me, since the router is connecting to a standby cluster (and we have yet to change/discuss if the router should connect to database through router to pick leadership changes on async cases)

paulomach:
> Thank you! LGTM. Q: do we still have usecases for clusteradmin OR we need to clean it one day in the future from Juju Secrets and Documentation?

`clusteradmin` it still in use, since it's the (less privileged) user used for group replication, just not for charm operation.

PR #520, opened 2024-09-11
DPE-5178 Upgrade fix for admin-address enabled charm

paulomach:
## Issue

When upgrading, the new charm will try to check for server "upgradability" using the admin_address/port of an unit not yet upgrade, i.e. it may not yet have the port open.

## Solution

Use `clusteradmin` user for the upgrade check, as it will default to standard port 3306.

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -134,7 +134,7 @@ LIBID = "8c1428f06b1b4ec8bf98b7d980a38a8c"
 # Increment this major API version when introducing breaking changes
 LIBAPI = 0
 
-LIBPATCH = 71
+LIBPATCH = 72
 
 UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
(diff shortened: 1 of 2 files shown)


shayancanonical (approved):

lib/charms/mysql/v0/mysql.py
```diff
@@ -134,7 +134,7 @@ def wait_until_mysql_connection(self) -> None:
 # Increment this major API version when introducing breaking changes
 LIBAPI = 0
 
-LIBPATCH = 71
+LIBPATCH = 72
```
taurus-forever, line 137: I read 72 was planned for @shayancanonical in some of reviews...

taurus-forever (approved):
LGTM, one comment.

lib/charms/mysql/v0/mysql.py
paulomach: It's the next one, so it depends who merges it first

PR #522, opened 2024-09-18
DPE-5493 Skip tls reconfiguration on unit teardown

paulomach:
## Issue

Caught on tests, uncaught exception on unit teardown when related to a TLS operator

## Solution

Skip TLS reconfiguration on unit teardown

### DIFF
--- a/lib/charms/mysql/v0/tls.py
+++ b/lib/charms/mysql/v0/tls.py
@@ -166,6 +166,9 @@ class MySQLTLS(Object):
 
     def _on_tls_relation_broken(self, _) -> None:
         """Disable TLS when TLS relation broken."""
+        if self.charm.removing_unit:
+            logger.debug("Unit is being removed, skipping TLS cleanup.")
+            return
         try:
             if not ops.jujuversion.JujuVersion.from_environ().has_secrets:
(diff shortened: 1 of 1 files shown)


lib/charms/mysql/v0/tls.py
```diff
@@ -174,6 +174,9 @@ def _on_tls_relation_broken(self, _) -> None:
         except KeyError:
             # ignore key error for unit teardown
             pass
+        if self.charm.removing_unit:
```
shayancanonical, line 177: im wondering if it might be a good idea to add this check to all event handlers (or better yet, condition all `self.framework.observe` statements to ensure no event handlers run if charm is being removed)

to better understand: why are we allowing setting of secrets on lines 170-173 instead of just no-oping the handler entirely

paulomach: yeah, eventually.

paulomach: > to better understand: why are we allowing setting of secrets on lines 170-173 instead of just no-oping the handler entirely

indeed. If unit is going away, so will it's databag keys

taurus-forever (approved):

PR #524, opened 2024-09-23
DPE-5540 Skip plugin install for not file found

paulomach:
## Issue

1. `config_change` hook will try to install plugins if configuration diff so indicates, even if plugins files are not available.
2. admin address enabled charms revisions will fail to connect to mysqld case it's not restarted

Case:
  `refresh` ran without `pre-upgrade-check` - so snap is not updated (no plugin files) but config changed is triggered.

## Solution
1. Check for plugin file before try install. Skip if file not found
2. Ensure mysqld restarts on post refresh `config_change` event

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -134,7 +134,7 @@ LIBID = "8c1428f06b1b4ec8bf98b7d980a38a8c"
 # Increment this major API version when introducing breaking changes
 LIBAPI = 0
 
-LIBPATCH = 73
+LIBPATCH = 74
 
 UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
 UNIT_ADD_LOCKNAME = "unit-add"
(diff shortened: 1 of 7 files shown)


shayancanonical (commented):
looks good. still confused about the symptoms of the issue - if the charm would enter an error state previously, i would recommend adding a test that runs upgrade without `pre-upgrade-check` to ensure that things start up normally. thoughts?

taurus-forever (approved):
LGTM here, ACK to Shayan: adding a test to run refresh without pre-upgrade-check will help us returning here for the following candidate revisions.

taurus-forever:
> ... if the charm would enter an error state previously...

Yes, see https://warthogs.atlassian.net/browse/DPE-5540

paulomach:
> looks good. still confused about the symptoms of the issue - if the charm would enter an error state previously, i would recommend adding a test that runs upgrade without `pre-upgrade-check` to ensure that things start up normally. thoughts?

You are right. I will include a test before merge

PR #527, opened 2024-09-28
DPE-4910 Change key logs from debug->info

paulomach:
## Issue

Some external hard to replicate bugs (e.g. #466 ) are even hard to debug to lack of useful logs.
(e.g. SQA tests are run with default info log level)

## Solution

Decreased log level to some key initialization and upgrade steps

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -244,6 +244,7 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
             BACKUPS_PASSWORD_KEY,
         ]
 
+        logger.info("Generating internal user credentials")
         for required_password in required_passwords:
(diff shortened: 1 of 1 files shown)


shayancanonical (approved):

src/charm.py
```diff
@@ -820,17 +819,16 @@ def join_unit_to_cluster(self) -> None:
                     from_instance=cluster_primary,
                     lock_instance=lock_instance,
                 )
-                logger.debug(f"Added instance {instance_address} to cluster")
```
taurus-forever, line 823: IMHO, it was useful message.

taurus-forever (approved):

src/charm.py
paulomach: it's redundant with the one at L831

PR #518, opened 2024-09-10
Port over charm lib changes from k8s charm relating to pod reschdule during cluster setup

shayancanonical:
Lib changes in https://github.com/canonical/mysql-k8s-operator/pull/499 to be published to charmhub

PR #528, opened 2024-09-30
DPE-5582 Timeout node count query

paulomach:
## Issue

Recent PR #518 introduced new method on update status for the k8s charm.
The method calls node_count method that fails on all units if there's one unresponsive unit, causing regression in freeze db tests, due responsive units not being able to relabel pods.

## Solution

Add timeout on node_count method to avoid hanging due unresponsive unit(s).


shayancanonical (approved):
LGTM!

taurus-forever (approved):

PR #532, opened 2024-10-04
DPE-5628 Skip set unknown config keys

paulomach:
## Issue

1. When rolling back to revisions with subset of known configuration, the unknown keys are wrongly tried to be dynamically set, which will break the rendered config dict.
2. Configuration is not always being persisted to disk

## Solution

1. Test config keys for existence in rendered config dict before trying to set them.
2. Always persist configuration to disk on config change

Tests to follow once this revision get promoted to stable

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -276,9 +276,6 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
             # the upgrade already restart the daemon
             return
 
-        # restart not required if mysqld is not running
-        mysqld_running = self._mysql.is_mysqld_running()
-
         previous_config = self.mysql_config.custom_config
         if not previous_config:
             # empty config means not initialized, skipping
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -298,30 +295,33 @@ def _on_config_changed(self, _) -> None:
...
-                self.on[f"{self.restart.name}"].acquire_lock.emit()
-                return
+            self.on[f"{self.restart.name}"].acquire_lock.emit()
 
-        if dynamic_config := self.mysql_config.filter_static_keys(changed_config):
+        elif dynamic_config := self.mysql_config.filter_static_keys(changed_config):
             # if only dynamic config changed, apply it
             logger.info("Configuration does not requires restart")
             for config in dynamic_config:
+                if config not in new_config_dict:
```
paulomach, line 322: these 3 lines are the main part

taurus-forever (approved):

PR #534, opened 2024-10-04
DPE-5415 VM reboot tests

paulomach:
## Issue

* missing vm host reboot tests

## Solution

* add test for rebooting unit hosts
* chore: tls test uses shared deploy method

### DIFF
--- a/tests/integration/test_vm_reboot.py
+++ b/tests/integration/test_vm_reboot.py
@@ -1,0 +1,84 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+import logging
+from pathlib import Path
+from time import sleep
+
+import pytest
+import yaml
+from pytest_operator.plugin import OpsTest
+
+from .high_availability.high_availability_helpers import (
+    deploy_and_scale_mysql,
+    get_application_name,
+)
+
+logger = logging.getLogger(__name__)
+
+
+METADATA = yaml.safe_load(Path("./metadata.yaml").read_text())
+
+
+@pytest.mark.group(1)
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test: OpsTest) -> None:
+    """Build the charm and deploy 3 units to ensure a cluster is formed."""
+    await deploy_and_scale_mysql(ops_test)
+
+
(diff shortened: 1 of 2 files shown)


src/charm.py
```diff
@@ -724,11 +724,19 @@ def _can_start(self, event: StartEvent) -> bool:
...
+            # empty config mean start never ran, skip next checks
+            return True
+
         # Safeguard if receiving on start after unit initialization
-        if self.unit_initialized:
-            logger.debug("Delegate status update for start handler on initialized unit.")
-            self._on_update_status(None)
-            return False
+        # with retries to allow time for mysqld startup
+        for _ in range(6):
```
paulomach, line 733: fix false negative on `self.unit_initialized` running before mysqld has a chance to start on reboot. If mysqld don't start within the 30s, recovery will happen on next update status

shayancanonical: may be we should use tenacity and raise an assertion error to keep retrying?

paulomach: i prefer to keep it simple since it is something to make recovery quicker

shayancanonical (approved):

tests/integration/test_vm_reboot.py
```diff
@@ -0,0 +1,97 @@
...
+        await ops_test.model.wait_for_idle(
+            apps=[app_name],
+            status="active",
+            timeout=15 * 60,
+        )
+
+
+def _machine_restart(machine_name: str) -> None:
+    """Restart the machine."""
+    run(["lxc", "restart", machine_name], check=True)
```
taurus-forever, line 97: Proposal. Can you please add: lxc stop --force machine_name.
This will stop processes unexpectedly. lxc retart will trigger systemd to shutdown DB gracefully (trigger primary switchover beforehand).

taurus-forever (approved):
LGTM with one proposal

tests/integration/test_vm_reboot.py
paulomach: @taurus-forever I say no on this test right now.

When doing the force stop for 2/3 or 3/3 causes loss of quorum, to which we still don't have a proper self-healing procedure, since doing it wrongly can cause transactions lost. Further discussion is required.

PR #543, opened 2024-10-23
DPE-5355 Ensure logs flushes

### LINKED ISSUE #523: logrotate config doesn't match layout on disk
## Steps to reproduce

1. Deploy (on jammy/amd64 here), and wait for enough workload to generate a lot of logs

## Expected behavior
logrotate should rotate mysql's log files to avoid filling the disk

## Actual behavior

logrotate looks for files that don't exist, and doesn't rotate the actual log files - which eventually will fill the disk.

logrotate config:
```
$ grep ^/var /etc/logrotate.d/flush_mysql_logs 
/var/snap/charmed-mysql/common/var/log/mysql/error.log {
/var/snap/charmed-mysql/common/var/log/mysql/general.log {
/var/snap/charmed-mysql/common/var/log/mysql/slowquery.log {
```

Actual contents of the log directory:
```
$ ls -l /var/snap/charmed-mysql/common/var/log/mysql/
total 268
drwxrwx--- 2 snap_daemon snap_daemon 122880 Sep 12 07:34 archive_error
drwxrwx--- 2 snap_daemon snap_daemon 139264 Sep 12 07:33 archive_general
drwxrwx--- 2 snap_daemon snap_daemon   4096 Sep 10 13:29 archive_slowquery
```

We found this file under the archive_general dir recently:
```
-rw-r----- 1 snap_daemon root 44G Sep 15 23:28 /var/lib/juju/storage/rootfs/30/29/var/log/mysql/archive_general/general.log-20240912_0733
```

## Versions

Operating system: Ubuntu 22.04.4 LTS

Juju CLI: 2.9.51-ubuntu-amd64

Juju agent: 2.9.49

Charm revision: 8.0/stable / 240

LXD: none

## Log output

Juju debug log: not relevant

## Additional context

The installed mysql snap is:
```
charmed-mysql        8.0.36         103    latest/stable  dataplatformbot  held
```

paulomach:
## Issue

Uncertain race/failure conditions can render a case where logs are being written to rotate file, i.e. the log file being written is under the `archive_<log>` directory.
Logrotate needs to match filename under the default directory to rotate AND flush the files.
This renders log rotation useless.

## Solution

* ensure file exist before rotation by touching it
* refactor for dynamic generating log rotation configuration
  * reacts to changes in configuration
* renamed slow query to align with mysql naming
* removed slow query from rotation list, since it's disable by default

Fixes #523


taurus-forever (approved):
Great simplification!

templates/run_log_rotation.sh.j2
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+LOG_PATH="{{ log_path }}"
```
sinclert-canonical, line 3: Why to define this env. variable? Does the `logrotate` process needs it?

src/mysql_vm_helpers.py
```diff
@@ -280,29 +280,48 @@ def write_mysqld_config(self) -> None:
...
+            logrotate_conf=config_path,
+            owner=MYSQL_SYSTEM_USER,
+            group=MYSQL_SYSTEM_USER,
         )
-        with open("/etc/cron.d/flush_mysql_logs", "w") as file:
-            file.write(cron)
+
+        self.write_content_to_file(script_path, logrotation_script_content, permission=0o550)
+
+        cron_content = f"* 1-23 * * * root {script_path}\n" f"1-59 0 * * * root {script_path}\n"
```
sinclert-canonical, line 323: Nit: could we have these two sub-strings in different lines somehow? I think they would be easier to read.

templates/run_log_rotation.sh.j2
paulomach: good catch
a6f3ad45

src/mysql_vm_helpers.py
paulomach: I just now notice it was 2 substr. Since it's pretty small, the formater (ruff) will rewrite as a single liner, and adding a line by, for instance putting in a tuple and using `"\n".join(substring_tuple)`, seems unnecessary this simple case.

sinclert-canonical: > Since it's pretty small, the formater (ruff) will rewrite as a single liner

Ugh, annoying. I wonder whether there is a rule to change that.

> Using "\n".join(substring_tuple) seems unnecessary this simple case.

Agreed.

sinclert-canonical (approved):

PR #550, opened 2024-11-19
Update snap_revisions.json with arm64

sinclert-canonical:
This PR adds architecture `armd64` to the list of supported ones within the [snap_revisions.json](https://github.com/canonical/mysql-operator/blob/70ba5f907cbf78c66561b37b5bf01cc90c0c4992/snap_revisions.json) file. Turns out different systems report their architectures differently. Some systems report it as `aarch64`, while others use `arm64`, depending on whether an Apple-based backend, or a LLVM based backend, is used ([reference](https://stackoverflow.com/a/47274698)).

Without this change, Apple laptops cannot be use for local development.


taurus-forever:
@sinclert-canonical What is your charmcraft version? 
I would check it with *ctaft team first, to follow their guide.

sinclert-canonical:
> I think aarch64 is exacly ARM64 case. Can you explain the reason for this commit?

Yes, my understanding is that both `aarch64` and `arm64` refer to the exact same architecture. After searching for it online, it seems one way of reporting it is due to Apple-based backend, while the other is due to community-based backend.

> What is your charmcraft version?

Not that is relevant for this problem, but my VM has `charmcraft` version `3.2.1` (latest).

> Also is it development (charmcraft pack) or deployment (juju deploy) issue?

No, I ran into this problem when executing tests on my Apple laptop:
```shell
$ tox run -e unit
> ...
> Traceback (most recent call last):
  File "/Users/<USERNAME>/Projects/Canonical/Data Platform/MySQL/mysql-operator/src/mysql_vm_helpers.py", line 183, in install_and_configure_mysql_dependencies
    revision = json.load(file)[platform.machine()]
               ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
KeyError: 'arm64'
```

sinclert-canonical:
Just realised the PostgreSQL operator repository has a similar structure to `snap-revisions` ([reference](https://github.com/canonical/postgresql-operator/blob/02367ec1fe8c714b2adc8d55a7dd65f14d92f1c4/src/constants.py#L34-L39)). Whatever we decide here, must also be applied there.

PR #552, opened 2024-11-21
[ONB-1301] Increase MySQLBase test coverage

sinclert-canonical:
This PR increases test coverage for the [MySQLBase class](https://github.com/canonical/mysql-operator/blob/70ba5f907cbf78c66561b37b5bf01cc90c0c4992/lib/charms/mysql/v0/mysql.py), by adding tests to 4 different untested methods.

Extending test coverage is one of the _Software Development_ onboarding tasks.

### DIFF
--- a/tests/unit/test_mysql.py
+++ b/tests/unit/test_mysql.py
@@ -873,7 +873,7 @@ class TestMySQLBase(unittest.TestCase):
         self.assertEqual(error.message, "Error message")
 
     @patch("charms.mysql.v0.mysql.MySQLBase._run_mysqlsh_script")
-    def test_delete_users_for_unit(self, _run_mysqlsh_script):
+    def test_delete_users_for_unit_failure(self, _run_mysqlsh_script):
         """Test failure to delete users for unit."""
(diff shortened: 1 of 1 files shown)


tests/unit/test_mysql.py
```diff
@@ -870,10 +873,15 @@ def test_error(self):
         self.assertEqual(error.message, "Error message")
 
     @patch("charms.mysql.v0.mysql.MySQLBase._run_mysqlsh_script")
-    def test_delete_users_for_relation_failure(
-        self,
-        _run_mysqlsh_script,
-    ):
+    def test_delete_users_for_unit(self, _run_mysqlsh_script):
```
shayancanonical, line 876: nit: should we add `_failure` suffix to indicate it is a test for failure?

shayancanonical (approved):

tests/unit/test_mysql.py
sinclert-canonical: Good catch! I see that suffix is also present in the test right below this one.

paulomach (approved):
Thank you

PR #555, opened 2024-11-26
[DPE-5711] Add warnings to destructive actions

sinclert-canonical:
This PR adds warning level messages when a _destructive_ action is performed. The definition of _destructive_ seems to be associated with data loss, which I think it can only happen, from user action, whenever they force a primary failover without checking considering the health of such instance.

### Additional info
- Paulo confirmed that this ticket was created to familiarised myself with the codebases.
- Alex referenced this PostgreSQL operator [PR](https://github.com/canonical/postgresql-k8s-operator/pull/753) as the reason behind the need for warnings.

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -1548,8 +1548,12 @@ class MySQLBase(ABC):
             ),
         )
 
-        try:
+        if force:
+            logger.warning(f"Promoting {cluster_name=} to primary with {force=}")
+        else:
             logger.debug(f"Promoting {cluster_name=} to primary with {force=}")
+
+        try:
             self._run_mysqlsh_script("\n".join(commands))
         except MySQLClientError:
             logger.exception("Failed to promote cluster to primary")
--- a/src/mysql_vm_helpers.py
+++ b/src/mysql_vm_helpers.py
@@ -831,6 +831,8 @@ class MySQL(MySQLBase):
     @staticmethod
     def reset_data_dir() -> None:
         """Reset the data directory."""
+        logger.warning(f"Resetting data directory: {MYSQL_DATA_DIR}")
+
         # Remove the data directory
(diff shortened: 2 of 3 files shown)


taurus-forever (approved):
re: `new libpatch?` = yes.

The changes are LGTM,

I would prepend with warning for sure:

src/mysql_vm_helpers.py
231:                shutil.rmtree(f"{CHARMED_MYSQL_DATA_DIRECTORY}/etc", ignore_errors=True)
835:        shutil.rmtree(MYSQL_DATA_DIR, ignore_errors=False)


I would also include the WARNING for calls like: 
> self._mysql.remove_instance in _on_database_storage_detaching() 
which can be triggered by juju on VM (planned reboot)... speculating here, but IMHO warning will not hard if cluster is loosing member.

Up 2 Paulo here.

sinclert-canonical:
> re: new libpatch? = yes.

👍🏻 

> I would prepend with warning for sure:
>
> src/mysql_vm_helpers.py
> 231: shutil.rmtree(f"{CHARMED_MYSQL_DATA_DIRECTORY}/etc", ignore_errors=True)
> 835: shutil.rmtree(MYSQL_DATA_DIR, ignore_errors=False)

Same as in [this comment](https://github.com/canonical/mysql-router-operator/pull/188#issuecomment-2504097691). Happy to do so if the criteria gets extended. Do you want me to?

sinclert-canonical:
Added warning to [MySQL.reset_data_dir](https://github.com/canonical/mysql-operator/blob/e04a142b2ffa4a8e91f29ab019e2158b1ca5c5f9/src/mysql_vm_helpers.py#L831-L841) method. [MySQL.uninstall_mysql](https://github.com/canonical/mysql-operator/blob/e04a142b2ffa4a8e91f29ab019e2158b1ca5c5f9/src/mysql_vm_helpers.py#L215-L233) one already have a log message with higher criticality (exception).

PR #558, opened 2024-11-27
fix: use juju provided ip (DPE-6105)

paulomach:
## Issue

On environments with multiple interfaces/networks, retrieving IP from a DNS may lead to unreachable address.

## Solution

Use juju provided bind address for hostname/ip map, instead of DNS returned address.

### DIFF
--- a/src/hostname_resolution.py
+++ b/src/hostname_resolution.py
@@ -50,14 +50,7 @@ class MySQLMachineHostnameResolution(Object):
         hostname = socket.gethostname()
         fqdn = socket.getfqdn()
 
-        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-        s.settimeout(0)
-        try:
-            s.connect(("10.10.10.10", 1))
-            ip = s.getsockname()[0]
-        except Exception:
-            logger.exception("Unable to get local IP address")
-            ip = "127.0.0.1"
+        ip = str(self.model.get_binding(PEER).network.bind_address)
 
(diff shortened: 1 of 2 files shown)


sinclert-canonical (approved):
👍🏻 LGTM.

A couple of questions, just for me to get a bit more context:
1. Previous implementation tried to connect to the socket, defaulting to 127.0.0.1. Are we OK trusting the model provided IP, without testing the connection?
2. There is another [snippet](https://github.com/canonical/mysql-operator/blob/90d679a153e968f13d5fccb2eb41c93538a68d58/src/ip_address_observer.py#L129-L140) within this repo with the same implementation. We cannot apply the same logic there as there is no `model` to be accessed, correct?

paulomach:
> 👍🏻 LGTM.
> 
> A couple of questions, just for me to get a bit more context:
> 
>     1. Previous implementation tried to connect to the socket, defaulting to 127.0.0.1. Are we OK trusting the model provided IP, without testing the connection?

We don't have much choice there. Testing can lead to a false negative, to which we don't have a safe escape hatch.

>     2. There is another [snippet](https://github.com/canonical/mysql-operator/blob/90d679a153e968f13d5fccb2eb41c93538a68d58/src/ip_address_observer.py#L129-L140) within this repo with the same implementation. We cannot apply the same logic there as there is no `model` to be accessed, correct?

Yes and this is used to detect address changes and trigger the update. Though the detection may give a false positive if, when having multiple interfaces, we may be testing against the wrong/unchanged one.

taurus-forever (approved):

PR #566, opened 2024-12-13
fix: [DPE-6234] Disable general_log and plugin audit_log_filter

taurus-forever:
## Issue

The config option general_log is not compatible with the real productions due to the massive SQL queries dump into the general.log: https://dev.mysql.com/doc/refman/8.0/en/query-log.html

> ... The server writes information to this log when clients connect or
> disconnect, and it logs each SQL statement received from clients...

In the past, it was a necessary evil as we had to keep all records about user's connects/disconnects for the security reasons. Nowadays we have audit_log plugin enabled by default, therefor no need to pay such a high performance price and struggle from the general.log size

https://github.com/canonical/mysql-operator/issues/523#issuecomment-2540596167
>  33GB /var/snap/charmed-mysql/common/var/log/mysql/archive_general/general.log-20241212_0013

Additionally, we are working on the logs compression spec (separate topic).

---

While working here, we have noticed unnecessary enabled plugin audit_log_filter which also produces log files in /var/lib/mysql. Disabling it for now, as it is not necessary for audit_log plugin announcer functionality.

We will consider re-enabling the plugin without rush in https://warthogs.atlassian.net/browse/DPE-6235

## Solution

* Permanently disable general_log (as no longer necessary)
* Temporary disable plugin audit_log_filter (to re-enable safely later).


shayancanonical (approved):

paulomach:
Close in favor of #574

PR #579, opened 2025-01-10
fix: DPE-6137 refator for password safety

paulomach:
## Issue

Passwords can be leaked in temporary files and cmdlines

## Solution

Refactor mysqlcli and mysqlsh calls to input password on stdin/console.

## Chore

* bumped mysql-connector-python version for integrations tests due to CVE
* fix tests


src/mysql_vm_helpers.py
```diff
@@ -733,49 +740,58 @@ def _run_mysqlsh_script(self, script: str, timeout=None) -> str:
...
-                    command, stderr=subprocess.PIPE, timeout=timeout
-                ).decode("utf-8")
+                output = subprocess.check_output(
+                    command,
+                    stderr=subprocess.PIPE,
+                    timeout=timeout,
+                    input=password,
+                    text=True,
+                    env=os.environ
+                    | {"MYSQL_TEST_LOGIN_FILE": f"{CHARMED_MYSQL_COMMON_DIRECTORY}/.mylogin.cnf"},
```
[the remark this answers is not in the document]
paulomach: It does not, as it would require the `--save-passwords=` option to be set and `--passwords-from-stdin` to be unset. This is undocumented behavior though, manually asserted.

charmcraft.yaml
```diff
@@ -37,3 +37,5 @@ parts:
       craftctl default
     charm-strict-dependencies: true
     charm-requirements: [requirements.txt]
+    charm-binary-python-packages:
+      - mysql-connector-python
```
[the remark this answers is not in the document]
paulomach: No, but I think it may be easy to build from source within the charm build.

The alternative to not use the lib would be to communicate with mysqlcli to input password, since it don't support read it from stdin.

lib/charms/mysql/v0/mysql.py
```diff
@@ -2399,24 +2578,27 @@ def get_member_state(self) -> Tuple[str, str]:
...
+        # [('ONLINE',
+        #  'PRIMARY',
+        #  '1de30105-ce16-11ef-bb27-00163e3cb985',
+        #  '1de30105-ce16-11ef-bb27-00163e3cb985'), (...)]
+        if len(output) == 0:
             raise MySQLNoMemberStateError("No member state retrieved")
 
-        if len(lines) == 2:
+        def lower_or_unknown(value) -> str:
+            return value.lower() if value else "unknown"
```
sinclert-canonical, line 2589: Should we check whether `value` is a `str` type, instead of whether it is Truthy? So far, this function is only used on CLI output text, but I am afraid its usages could be expand in the future.

```python
return value.lower() if isinstance(value, str) else "unknown"
```

pyproject.toml
```diff
@@ -15,6 +15,7 @@ pyopenssl = "^24.0.0"
 typing_extensions = "^4.7.1"
 jinja2 = "^3.1.2"
 python_hosts = "^1.0.6"
+pexpect = "^4.9.0"
```
sinclert-canonical, line 18: Is this dependency really necessary? It seems it is only used once (see [code](https://github.com/canonical/mysql-operator/blob/bd92684fc86bdb354623575b5341f10d021f15bc/src/mysql_vm_helpers.py#L780-L808)).

src/mysql_vm_helpers.py
```diff
@@ -703,56 +700,68 @@ def restart_mysql_exporter(self) -> None:
...
+        try:
+            output = subprocess.check_output(
+                command,
+                stderr=subprocess.PIPE,
+                timeout=timeout,
+                input=password,
+                text=True,
+            )
+            # split output to clean mysqlsh garbage
+            return output.split("###")[1].strip()
```
sinclert-canonical, line 744: I understand we are removing a `###` prefix from the output, with `output` being similar to [this example](https://github.com/canonical/mysql-operator/blob/bd92684fc86bdb354623575b5341f10d021f15bc/tests/unit/test_mysqlsh_helpers.py#L74) from the test suite. However, should we use Python 3.9+ [`removeprefix`](https://docs.python.org/3/library/stdtypes.html#str.removeprefix) function to better express the intent? 🤔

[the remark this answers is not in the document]
paulomach: We will get there

pyproject.toml
paulomach: Yes. That's the lib used to interact with mysql cli since it cannont input password from stdin, but it does from the console

src/mysql_vm_helpers.py
```diff
@@ -703,56 +700,68 @@ def restart_mysql_exporter(self) -> None:
...
-
-            command = [
-                CHARMED_MYSQLSH,
-                "--no-wizard",
-                "--python",
-                "-f",
-                _file.name,
-            ]
+        # prepend every shell command with
+        # - set wizard to false. cannot be set on cmd line option as it conflicts with --passwords-from-stdin
```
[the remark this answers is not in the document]
paulomach: If `--no-wizard` is set, stdin password will fail:
```
root@noble:~# echo newpass | mysqlsh -u root --passwords-from-stdin --pyc "print('I am logged')"
ERROR: Failed to retrieve the password: Process "mysql_config_editor" finished with exit code 1: mysql_config_editor: Can't create/write to file '/root/snap/charmed-mysql/x1/.mylogin.cnf' (OS errno 13 - Permission denied)
couldn't create the login file : Permission denied
mysql_config_editor: Error on close of '<fd out of range>' (OS errno 9 - Bad file descriptor)
operation failed. : Bad file descriptor
Please provide the password for 'root@localhost': I am logged
root@noble:~# echo newpass | mysqlsh -u root --nw  --passwords-from-stdin --pyc "print('I am logged')"
ERROR: Failed to retrieve the password: Process "mysql_config_editor" finished with exit code 1: mysql_config_editor: Can't create/write to file '/root/snap/charmed-mysql/x1/.mylogin.cnf' (OS errno 13 - Permission denied)
couldn't create the login file : Permission denied
mysql_config_editor: Error on close of '<fd out of range>' (OS errno 9 - Bad file descriptor)
operation failed. : Bad file descriptor
MySQL Error 1045: Access denied for user 'root'@'localhost' (using password: NO)
```

We still set no-wizard after login so interactive commands go (e.g. configure_instance).

I haven't tested everything through stdin, but then I guess it's a matter of taste

src/mysql_vm_helpers.py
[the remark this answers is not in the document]
paulomach: eventually

src/mysql_vm_helpers.py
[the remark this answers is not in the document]
paulomach: I tested it. I think I prefer as is because it gives the failed command directly on the traceback, since it's will be on the cmd line

src/mysql_vm_helpers.py
sinclert-canonical: Same as [this comment](https://github.com/canonical/mysql-k8s-operator/pull/553#discussion_r1916330533).

paulomach: 13278c34

lib/charms/mysql/v0/mysql.py
paulomach: When the need arrives we change it. For the specific case, it can return as an empty string, which evaluates to `False`

taurus-forever (approved):

PR #581, opened 2025-01-13
Separate IP address observer into separate script

carlcsaposs-canonical:
Before, ip_addresss_observer.py simultaneously solved two different concerns, one to provide imports used in charm.py and one to call charm.py itself (via juju-run).

Now, those two concerns are separated.

Fixes issue where ops import caused `juju-run` script to fail because ops is not available from /usr/bin/python3 when using a full virtual environment (needed for charmcraft 3 poetry plugin migration)—now `juju-run` script does not import ops since it never needed it

Context: https://chat.canonical.com/canonical/pl/drhbrt84y3yc7xwc5431a1o5de

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -19,6 +19,8 @@ parts:
       - -requirements.txt
     prime:
       - snap_revisions.json
+      - scripts
+      - templates
   charm:
     build-snaps:
       - rustup
--- a/scripts/ip_address_dispatcher.py
+++ b/scripts/ip_address_dispatcher.py
@@ -1,0 +1,59 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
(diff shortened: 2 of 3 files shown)


charmcraft.yaml
```diff
@@ -19,6 +19,8 @@ parts:
       - -requirements.txt
     prime:
       - snap_revisions.json
+      - scripts
+      - templates
```
paulomach, line 23: :thinking: wasn't templates there already!

paulomach (approved):
lgtm. Just waiting for the tests on the test branch

sinclert-canonical (approved):

charmcraft.yaml
taurus-forever: Wow... how did the logrotate work without logrotate templates in prime? o_0

taurus-forever (approved):

PR #589, opened 2025-01-17
Remove pyopenssl

carlcsaposs-canonical:
Not sure why we added this in the first place; given that if pyopenssl is importable, urllib3 will use insecure cryptography


shayancanonical (approved):

lib/charms/mysql/v0/architecture.py
```diff
@@ -41,8 +41,6 @@
 LIBAPI = 0
 LIBPATCH = 1
 
-PYDEPS = ["ops>=2.0.0", "pyyaml>=5.0"]
```
sinclert-canonical, line 44: Why are we removing this? The PR description does not cover it.

AFAIK, charm libs should specify the list of packages that they expect in their environment, so later on charmed applications add them to their list of dependencies. These two (`ops` and `pyyaml`) are indeed used within the [architecture.py](https://github.com/canonical/mysql-operator/blob/main/lib/charms/mysql/v0/architecture.py) charm lib.

sinclert-canonical: This one is shared with the `mysql-k8s-operator` charm, even if that one also belong to the team.

I am happy with whatever you guys decide, but when I learned about charm libs I saw no distinction between externally and internally shared ones. A charm lib is a lib, and therefore it must specify its dependencies _some-how_.

sinclert-canonical: Not an approach that I like (i.e. not stating dependencies on a _lib_ type of software artefact), but I realize that it is, indeed, the approach you guys have followed so far.

I will approve the changes

sinclert-canonical (approved):

paulomach (approved):

PR #592, opened 2025-01-21
[DPE-4375] Add cluster manual re-join handler

sinclert-canonical:
This PR adds logic to manually re-join a MySQL replica that has gone `OFFLINE`, to the cluster, whenever MySQL 8.0.21+ [auto re-join](https://dev.mysql.com/doc/refman/8.0/en/group-replication-responses-failure-rejoin.html) attempts have been exhausted.

### Description

There are edge cases where a MySQL instance that has lost connection to the cluster it originally belong to (i.e. `OFFLINE`), exhaust its automatic retries (by default 3 retries, with 5 mins between each, starting with MySQL 8.0.21). For those cases, a manual re-join should be performed.

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -133,7 +133,7 @@ LIBID = "8c1428f06b1b4ec8bf98b7d980a38a8c"
 # Increment this major API version when introducing breaking changes
 LIBAPI = 0
 
-LIBPATCH = 80
+LIBPATCH = 81
 
 UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
 UNIT_ADD_LOCKNAME = "unit-add"
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -2605,6 +2605,31 @@ class MySQLBase(ABC):
 
         raise MySQLNoMemberStateError("No member state retrieved")
 
+    def is_cluster_auto_rejoin_ongoing(self):
+        """Check if the instance is performing a cluster auto rejoin operation."""
+        cluster_auto_rejoin_command = (
+            "cursor = session.run_sql(\"SELECT work_completed, work_estimated FROM performance_schema.events_stages_current WHERE event_name LIKE '%auto-rejoin%'\")",
+            "result = cursor.fetch_one() or [0,0]",
+            "print(f'<COMPLETED_ATTEMPTS>{result[0]}</COMPLETED_ATTEMPTS>')",
+            "print(f'<ESTIMATED_ATTEMPTS>{result[1]}</ESTIMATED_ATTEMPTS>')",
+        )
+
+        try:
+            output = self._run_mysqlsh_script(
+                "\n".join(cluster_auto_rejoin_command),
+                user=self.server_config_user,
+                password=self.server_config_password,
+                host=self.instance_def(self.server_config_user),
+            )
+        except MySQLClientError as e:
+            logger.error("Failed to get cluster auto-rejoin information", exc_info=e)
+            raise
+
+        completed_matches = re.search(r"<COMPLETED_ATTEMPTS>(\d)</COMPLETED_ATTEMPTS>", output)
+        estimated_matches = re.search(r"<ESTIMATED_ATTEMPTS>(\d)</ESTIMATED_ATTEMPTS>", output)
+
+        return int(completed_matches.group(1)) < int(estimated_matches.group(1))
+
     def is_cluster_replica(self, from_instance: Optional[str] = None) -> Optional[bool]:
         """Check if this cluster is a replica in a cluster set."""
         cs_status = self.get_cluster_set_status(extended=0, from_instance=from_instance)
--- a/src/charm.py
+++ b/src/charm.py
@@ -459,6 +459,14 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
                 except MySQLRebootFromCompleteOutageError:
                     logger.error("Failed to reboot cluster from complete outage.")
                     self.unit.status = BlockedStatus("failed to recover cluster.")
+                finally:
+                    return
+
+            if self._mysql.is_cluster_auto_rejoin_ongoing():
+                logger.info("Cluster auto-rejoin attempts are still ongoing.")
+            else:
+                logger.info("Cluster auto-rejoin attempts are exhausted. Attempting manual rejoin")
+                self._execute_manual_rejoin()
 
         if state == "unreachable":
             try:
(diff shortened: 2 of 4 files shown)


lib/charms/mysql/v0/mysql.py
```diff
@@ -2420,6 +2420,27 @@ def get_member_state(self) -> Tuple[str, str]:
 
         raise MySQLNoMemberStateError("No member state retrieved")
 
+    def is_cluster_auto_rejoin_ongoing(self):
+        """Check if the instance is performing a cluster auto rejoin operation."""
+        cluster_auto_rejoin_command = (
+            f"shell.connect('{self.instance_def(self.server_config_user)}')",
+            "cursor = session.run_sql('SELECT WORK_COMPLETED, WORK_ESTIMATED FROM performance_schema.events_stages_current WHERE EVENT_NAME LIKE '%auto-rejoin%'')"
+            "result = cursor.fetch_one() or [0,0]"
```
sinclert-canonical, line 2428: The `[0.0]` value is a safe default, to actually differentiate between the case where `cursor.fetch_one()` returns None, from the case where the SQL query actually crashes (agreed with Paulo online).

src/charm.py
```diff
@@ -460,6 +460,13 @@ def _handle_non_online_instance_status(self, state) -> None:
                     logger.error("Failed to reboot cluster from complete outage.")
                     self.unit.status = BlockedStatus("failed to recover cluster.")
 
+            elif self._mysql.is_cluster_auto_rejoin_ongoing():
+                logger.info("Cluster auto-rejoin attempts are still ongoing.")
+
+            elif not self._mysql.is_cluster_auto_rejoin_ongoing():
+                logger.info("Cluster auto-rejoin attempts are exhausted. Attempting manual rejoin")
+                self._execute_manual_rejoin()
```
sinclert-canonical, line 468: Both these `elif` clauses had a `not self.is_leader()` check before. However, Paulo confirmed that there are scenarios where a replica can end up as a Juju leader, and the manual re-join logic must affect those too.

lib/charms/mysql/v0/mysql.py
```diff
@@ -2420,6 +2420,27 @@ def get_member_state(self) -> Tuple[str, str]:
 
         raise MySQLNoMemberStateError("No member state retrieved")
 
+    def is_cluster_auto_rejoin_ongoing(self):
+        """Check if the instance is performing a cluster auto rejoin operation."""
+        cluster_auto_rejoin_command = (
+            f"shell.connect('{self.instance_def(self.server_config_user)}')",
+            "cursor = session.run_sql(\"SELECT work_completed, work_estimated FROM performance_schema.events_stages_current WHERE event_name LIKE '%auto-rejoin%'\")",
+            "result = cursor.fetch_one() or [0,0]",
```
sinclert-canonical, line 2428: The `[0.0]` value is a safe default, to actually differentiate between the case where `cursor.fetch_one()` returns None, from the case where the SQL query actually crashes (agreed with Paulo online).

sinclert-canonical: I originally had `return False` whenever the query crashed. However, Paulo told me to differentiate between _expected_ and _unexpected_ crashes, to avoid noise within the logs. So we decided to `raise` the exception instead.

You are right that, after such change, the log level should probably be `ERROR`.

src/charm.py
```diff
@@ -449,7 +449,7 @@ def _handle_non_online_instance_status(self, state) -> None:
             }
             all_states.add("offline")
 
-            if all_states == {"offline"} and self.unit.is_leader():
+            if self.unit.is_leader() and all_states == {"offline"}:
```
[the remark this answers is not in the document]
sinclert-canonical: Honestly none.

I originally altered the order of the conditions looking for consistency with the other `elsif` clauses, when I had ~similar checks there (see [this comment](https://github.com/canonical/mysql-operator/pull/592#discussion_r1924950890)). Now that those conditionals were removed, this change is useless.

src/charm.py
```diff
@@ -460,6 +460,13 @@ def _handle_non_online_instance_status(self, state) -> None:
                     logger.error("Failed to reboot cluster from complete outage.")
                     self.unit.status = BlockedStatus("failed to recover cluster.")
 
+            elif self._mysql.is_cluster_auto_rejoin_ongoing():
+                logger.info("Cluster auto-rejoin attempts are still ongoing.")
+
+            elif not self._mysql.is_cluster_auto_rejoin_ongoing():
```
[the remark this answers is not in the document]
sinclert-canonical: You have a point about race conditions, but I believe they can be solve with a simple `return`.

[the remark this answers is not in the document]
sinclert-canonical: Not sure I follow.

The proposed [is_cluster_auto_rejoin_ongoing](https://github.com/canonical/mysql-operator/blob/sinclert/4375/cluster-manual-rejoin/lib/charms/mysql/v0/mysql.py#L2423-L2442) function only fetches info from the `events_stages_current` MySQL table. It does not update it in any way, nor it triggers the auto re-join procedure.

Alternatively, the [_execute_manual_rejoin](https://github.com/canonical/mysql-operator/blob/sinclert/4375/cluster-manual-rejoin/src/charm.py#L481-L503) function performs a set of actions that are not registered within the `events_stages_current` MySQL table, at least not containing the _auto-rejoin_ string in their event names.

lib/charms/mysql/v0/mysql.py
sinclert-canonical: Changed in commit https://github.com/canonical/mysql-operator/pull/592/commits/bd8292d5dac50d15e9031502e8bda9c29504cbc1.

src/charm.py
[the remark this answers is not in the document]
sinclert-canonical: Reverted in commit https://github.com/canonical/mysql-operator/pull/592/commits/bd8292d5dac50d15e9031502e8bda9c29504cbc1.

lib/charms/mysql/v0/mysql.py
sinclert-canonical: Exactly, that is _expected_, given that we could be checking for _auto-rejoin_ event-names at moments where there are none. The error logging is for those _unexpected_ cases where the query crashes, somehow (i.e. future versions of MySQL rename the table columns, the MySQL is saturated and times out, etc).

sinclert-canonical: > so if it returns None the query was successful?

Yes. That is a perfectly valid, though disappointed, result 😆

paulomach (commented):
@sinclert-canonical What's the idea for a integration test?

sinclert-canonical:
Branch rebased from `main`.

sinclert-canonical:
Branch rebased from `main`.

paulomach (approved):
Good stuff

shayancanonical (approved):

shayancanonical (approved):

PR #597, opened 2025-01-24
feat: DPE-5656 log rotation new options

### LINKED ISSUE #539: Log level and log retention ought to be operator-configurable
## Steps to reproduce

1. Deploy a busy unit on a disk-limited flavor - say, 50GB of disk space for a VM running a 5GB database.
2. Throw enough traffic at the application driven by that database for a lot of logs to be written
3. Watch the disk fill up

## Expected behavior

The operator should be able to prevent the disk from filling up by reducing either the log level, or the log retention, or a combination of both. Enabling compression for older logs may also be something to (re)consider, at least as a non-default option.

## Actual behavior

If the unit is sufficiently busy (say, a content database for a popular data-driven website), the logs will quickly fill up the disk.

We've seen numbers up to the region of 1GB/hr of logging against a 5GB DB, so best part of 25GB of logging per day - with 7 days retention hard-coded, that's potentially 175GB or so of logs on disk at any time, about 35x the size of the database itself.

This requirement may not become clear to less attentive operators until up to a week after deployment time, or may creep up if the application using the database gradually sees more and more traffic over time.

## Versions

Operating system: Ubuntu 22.04.4 LTS

Juju CLI: 2.9.50-ubuntu-amd64

Juju agent: 2.9.49

Charm revision: 240

LXD: N/A

## Log output

Juju debug log: N/A

## Additional context

N/A

paulomach:
## Issue

[DA124 implementation](https://docs.google.com/document/d/1KtI2_eDI0srjG6rGKUzxGpEJ2uvBDYhgKN_6-rIjMJE/edit?tab=t.0)

## Solution
Adds:
* new config option for log retention period
* new config option for audit log policy
* dynamic behavior for log compression and log retention period, depending on COS/logs being pushed.

Fixes #539


scripts/wait_for_log_sync.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+#
+# Trigger custom juju event when promtail start syncing logs
+#
+
+CHARM_DIR="/var/lib/juju/agents/unit-$(echo "${JUJU_UNIT_NAME}" | tr / -)/charm"
+POSITIONS_FILE="/var/snap/grafana-agent/current/grafana-agent-positions/log_file_scraper.yml"
+
+# Support bin path for 3.x and 2.x
+[[ -x "/usr/bin/juju-exec" ]] && JUJU_CMD="/usr/bin/juju-exec" || JUJU_CMD="/usr/bin/juju-run"
```
sinclert-canonical, line 10: Is this some kind of `JUJU_CMD = if... else...` assignment?

src/charm.py
```diff
@@ -284,25 +287,19 @@ def _on_config_changed(self, _) -> None:
...
-            path=MYSQLD_CUSTOM_CONFIG_FILE, content=new_config_content
-        )
-        self._mysql.setup_logrotate_and_cron(self.text_logs)
+        # Log rotation setup as DA124
+        if self.config.logs_retention_period == "auto" and self.model.get_relation(
+            COS_AGENT_RELATION_NAME
+        ):
+            retention_period, compress = "1", self.unit_peer_data.get("logs_synced") == "true"
+        else:
+            retention_period, compress = "3", True
```
sinclert-canonical, line 300: Any specific reason why assignments are done in-line here? Technically works, but I think it is fairly uncommon throughout the codebase.

src/log_rotation_setup.py
```diff
@@ -0,0 +1,87 @@
...
+            self.charm.on[COS_AGENT_RELATION_NAME].relation_broken, self._cos_relation_broken
+        )
+
+    def _cos_relation_created(self, _):
+        """Start monitoring script that tracks start of log syncing."""
+        script_path = f"{self.charm.charm_dir}/scripts/wait_for_log_sync.sh"
+
+        new_env = os.environ.copy()
+        if "JUJU_CONTEXT_ID" in new_env:
+            new_env.pop("JUJU_CONTEXT_ID")
```
sinclert-canonical, line 54: Why is this sanitisation required? Is there any other env. variable you expect could be conflicting in the future?

src/mysql_vm_helpers.py
```diff
@@ -277,18 +278,34 @@ def write_mysqld_config(self) -> None:
...
         logger.debug("Creating logrotate config file")
         config_path = "/etc/logrotate.d/flush_mysql_logs"
         script_path = f"{self.charm.charm_dir}/logrotation.sh"
         cron_path = "/etc/cron.d/flush_mysql_logs"
         logs_dir = f"{CHARMED_MYSQL_COMMON_DIRECTORY}/var/log/mysql"
 
+        if logs_retention_period == "auto":
+            retention_period = 3
+        else:
+            retention_period = int(logs_retention_period)
```
sinclert-canonical, line 305: Could we make this function to receive `logs_retention_period` already as an integer?

I am under the impression that the "auto" config option can only be set via charm config, and the charm is already translating it to integers [here](https://github.com/canonical/mysql-operator/blob/50f5f4ad24ff062ee60a9680c3d12518ace91f58/src/charm.py#L294-L300). We may need to tweak the logic a little bit, to make the "auto" parsing COS-relation independent.

sinclert-canonical (commented):
Initial code revision, requested by Paulo.

src/log_rotation_setup.py
sinclert-canonical: Got it, thanks! I agree a comment will help, this is implicit knowledge about the runtime itself (Juju)

config.yaml
```diff
@@ -44,6 +44,18 @@ options:
     description: Number of days for binary logs retention
     type: int
     default: 7
+  logs_audit_policy:
+    description: |
+      Audit log policy. Allowed values are: "all", "logins" (default), "queries".
+      Reg. at https://docs.percona.com/percona-server/8.0/audit-log-plugin.html#audit_log_policy
```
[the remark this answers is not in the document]
paulomach: Although the config looks the same, we do pack the percona version of the audit log and not the mysql enterprise one.

scripts/wait_for_log_sync.sh
paulomach: yes - aka "bashism"

src/charm.py
paulomach: no special reason, can change

config.yaml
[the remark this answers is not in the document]
sinclert-canonical: Ok, feel free to leave it as is if there is a reason to point folks to those docs instead.

src/mysql_vm_helpers.py
paulomach: changed

sinclert-canonical (approved):
Looks great. Thanks for addressing all my comments!

config.yaml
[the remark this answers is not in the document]
paulomach: good catch, it's a typo to "Ref" of reference

config.yaml
```diff
@@ -44,6 +44,18 @@ options:
...
+    type: string
+    default: logins
+  logs_retention_period:
+    description: |
+      Period for rotated logs retention, in days. When set to "auto" (default) it will default
+      to 3 days, or 1 day if related to COS. Accept values equal or greater then 3 otherwise.
```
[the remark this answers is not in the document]
paulomach: You understood correctly and will not accept 1 in the config (set to "1" is dynamic only). What do you suggest?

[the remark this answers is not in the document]
paulomach: How about:
"Specifies the retention period for rotated logs, in days. Accepts an integer value of 3 or greater, or the special value "auto". When set to "auto" (default), the retention period is 3 days, except when COS-related, where it is 1 day."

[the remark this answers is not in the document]
paulomach: dfb56311a

config.yaml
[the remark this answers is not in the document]
paulomach: dfb56311a

paulomach:
> The log-syncing script once a COS relation is formed that waits to ensure that logs are being synced before changing the logrotation config -> the whole mechanism seems clunky. However, it seems like a sure fire way of ensuring that logs are syncing before the logrotation config is changed
> 
> Another option would be to check `grep mysql.*log ${POSTIONS_FILE}` in the constructor of the charm (along with a no-op if the action is already taken) -> this will be unexpected-error proof with a retry mechanism built in (when the next update-status runs). Thoughts?

On point. I've refactored to remove the need of the custom event, and rely on update_status instead, since this can be somewhat lazy

PR #604, opened 2025-01-31
[MISC] Fix set-dynamic-variable lib function

sinclert-canonical:
This PR aims to fix a hidden bug related to the setting of MySQL variables, where the variable names where not escaped correctly, and the regex syntax was misleading (no need to put the `,` character between ranges). This bug was discovered while checking passing Upgrade integration tests (see [CI run](https://github.com/canonical/mysql-operator/actions/runs/13055730889/job/36429441154?pr=603#step:28:264))


paulomach:
@sinclert-canonical , this fix is not valid. This dict contains config file keys that aren't possible to set at runtime, i.e. requires a daemon restart. Which is not the case for some of the keys you added like [audit log policy](https://docs.percona.com/percona-server/8.0/audit-log-plugin.html#audit_log_policy)

An appropriated fix is proposed on the [PR597](https://github.com/canonical/mysql-operator/pull/597/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR31), line 317

sinclert-canonical:
> @sinclert-canonical , this fix is not valid. This dict contains config file keys that aren't possible to set at runtime, i.e. requires a daemon restart. Which is not the case for some of the keys you added like [audit log policy](https://docs.percona.com/percona-server/8.0/audit-log-plugin.html#audit_log_policy)

For anyone arriving here from the future, the confusing came from:
- Paulo: using the Percona MySQL docs (see [audit_log_policy](https://docs.percona.com/percona-server/8.0/audit-log-plugin.html#audit_log_policy) spec).
- Sinclert: using the official MySQL docs (see [audit_log_policy](https://dev.mysql.com/doc/refman/8.0/en/audit-log-reference.html#sysvar_audit_log_policy) spec).

> An appropriated fix is proposed on the [PR597](https://github.com/canonical/mysql-operator/pull/597/files#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR31), line 317

Not entirely, as the upgrade integration tests keep crashing silently (see [CI run](https://github.com/canonical/mysql-operator/actions/runs/13052381304/job/36420350830?pr=597#step:28:260)). Further investigation is needed.

sinclert-canonical:
PR https://github.com/canonical/mysql-operator/pull/597 super-seeds this one by removing the `loose-` prefix from any MySQL variable, and removing `audit_log_filter` variable from the repository.

PR #607, opened 2025-02-04
fix: DPE-6485 optional restart on instance configuration

paulomach:
## Issue

Instance configuration need to expose `restart` for explicit restart under pebble.

## Solution

* Expose `restart`
* refactor restart method to ditch rolling ops event defer and reuse unit recovery method (from upgrade code)


shayancanonical (approved):

sinclert-canonical (approved):

src/charm.py
```diff
@@ -873,28 +873,50 @@ def join_unit_to_cluster(self) -> None:
...
-            # defer restart until all units are in the relation
-            logger.debug("Deferring restart until all units are in the relation")
-            event.defer()
+    def recover_unit_after_restart(self) -> None:
+        """Wait for unit recovery/rejoin after restart."""
+        recovery_attempts = 30
+        logger.info("Recovering unit")
+        if self.app.planned_units() == 1:
+            self._mysql.reboot_from_complete_outage()
+        else:
```
sinclert-canonical, line 882: A small improvement to reduce indentation levels here would be to add a `return` clause within the `if` statement, and leave everything within the `else` block un-indented.

lib/charms/mysql/v0/mysql.py
```diff
@@ -1955,6 +1961,7 @@ def is_instance_in_cluster(self, unit_label: str) -> bool:
                 user=self.server_config_user,
                 password=self.server_config_password,
                 host=self.instance_def(self.server_config_user),
+                exception_as_warning=True,
```
taurus-forever, line 1964: Can you please share some explanation here? Tnx!

taurus-forever (approved):
LGTM, one curiosity question.

lib/charms/mysql/v0/mysql.py
paulomach: When the method is called on an instance that does not belong to the cluster (initialization, some self healing cases), it throws an noisy error log which is not a failure per se, but a valid output.

PR #609, opened 2025-02-06
Use `charmcraft test` & concierge

carlcsaposs-canonical:
Ported from https://github.com/canonical/mysql-router-k8s-operator/pull/379

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -13,7 +13,7 @@ jobs:
     uses: ./.github/workflows/ci.yaml
     secrets: inherit
     permissions:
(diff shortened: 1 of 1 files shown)


sinclert-canonical (commented):
Same comments as in https://github.com/canonical/mysql-router-operator/pull/210.

taurus-forever (approved):
LGTM to unblock mid-sprint delivery. 
Possible CI/CD improvements (including readability and python scripts extractions will follow as a separate PRs).

paulomach (approved):
I would prefer to split tests using the test name, e.g. `s/test_replication_1/test_exporter_endpoints/` 
That can be done later though

PR #615, opened 2025-02-13
[DPE-6488] Address slow mysqld start upon unit reboot

shayancanonical:
## Issue
In vsphere models, mysqld sometimes takes a long time to start up when a unit is rebooted. In these cases, we wrongly assume that the cluster is not already initialized and try to initialize the cluster. Instead, what we should do is defer the start event if mysqld does not come up within a fixed amount of time

## Solution
- Refactor `unit_initialized` and `get_cluster_metadata` to bubble up errors if mysqld client errors out (due to timeout or similar reasons). The refactor will use mysqlcli to get cluster metadata locally and mysqlsh to connect and get cluster metadata from remote units
- Convert `check_mysqlsh_connection` -> `check_mysqlcli_connection`, and use mysqlcli to `get_cluster_metadata` if checking in local unit (since after reboot in vsphere, admin port may not be already be open causing mysqlsh connections against 33062/admin-port to timeout. after a reboot in vsphere, mysqld is restarted because /etc/hosts is modified to drop loopback ip addresses -> during this time, the admin-port may timeout)

## Additional notes

- `check_mysqlcli_connection` is refactored to use `serverconfig` credentials (since these credentials will exist always except the first run when the user is yet to be created). if for the first time, `serverconfig` credentials dont exist, then fallback to checking with `root` user and the unix socket


lib/charms/mysql/v0/mysql.py
```diff
@@ -1674,35 +1685,63 @@ def is_cluster_in_cluster_set(self, cluster_name: str) -> Optional[bool]:
...
-            "  print(bool(result.fetch_one()))",
-            "else:",
-            "  print(False)",
+        if from_instance:
+            check_cluster_metadata_commands = (
+                "result = session.run_sql(\"SHOW DATABASES LIKE 'mysql_innodb_cluster_metadata'\")",
+                "content = result.fetch_all()",
+                "if content:",
+                (
+                    '  result = session.run_sql("SELECT cluster_name FROM mysql_innodb_cluster_metadata'
```
shayancanonical, line 1696: possibly use the same `select from where where exists` as below

shayancanonical: addressed in https://github.com/canonical/mysql-operator/pull/615/commits/bded9d5176239dbe6d93df32eba88049aa853637

src/hostname_resolution.py
```diff
@@ -89,23 +89,29 @@ def _get_host_details(self) -> list[HostsEntry]:
...
             logger.debug("No hostnames in the peer databag. Skipping update of /etc/hosts")
-            return
+            return False
 
         logger.debug("Updating /etc/hosts with new hostname to IP mappings")
         hosts = Hosts()
 
+        loopback_host_exists = False
+
         if hosts.exists(address="127.0.1.1", names=[socket.getfqdn()]):
```
paulomach, line 110: summon the walrus!

```suggestion
        if loopback_host_exists := hosts.exists(address="127.0.1.1", names=[socket.getfqdn()]):
```

src/mysql_vm_helpers.py
```diff
@@ -830,7 +832,8 @@ def _run_mysqlcli_script(
                 if len(stdout) > 1 and "ERROR" in stdout[1]:
                     # some errors will not bubble up from spawned process
                     # but are reported in the stdout
-                    logger.error(stdout[1].strip())
+                    if log_errors:
```
paulomach, line 835: What's the error you wanted to show?

paulomach (commented):
Thoughts on having the treatment at the `on_start` hook?

I tested thoroughly and it works nicely, though more complex than I initially thought it would be.

src/mysql_vm_helpers.py
shayancanonical: it's actually an error that i did not want to log to the debug log (especially since cluster metadata will not exist before a unit joins the cluster - we will unnecessarily log errors)

lib/charms/mysql/v0/mysql.py
```diff
@@ -1674,35 +1689,63 @@ def is_cluster_in_cluster_set(self, cluster_name: str) -> Optional[bool]:
...
-            "  print(False)",
+        if from_instance:
+            check_cluster_metadata_commands = (
+                'cursor = session.run_sql("SELECT cluster_name '
+                "FROM mysql_innodb_cluster_metadata.clusters "
+                "WHERE EXISTS ("
+                "SELECT * "
+                "FROM information_schema.schemata "
+                "WHERE schema_name = 'mysql_innodb_cluster_metadata'"
+                ')")',
```
sinclert-canonical, line 1702: This query is the same as [the one](https://github.com/canonical/mysql-operator/blob/5a82aa4e888b04f573c1f23f663aa9a70b0a07ee/lib/charms/mysql/v0/mysql.py#L1724-L1731) defined a few lines below.

Could we extract it to a constant variable? If you don't think it necessary, could we at least write them with the same format (i.e. line splits at the same positions).

lib/charms/mysql/v0/mysql.py
```diff
@@ -1674,35 +1689,63 @@ def is_cluster_in_cluster_set(self, cluster_name: str) -> Optional[bool]:
...
+
+        check_cluster_metadata_query = (
+            "SELECT cluster_name "
+            "FROM mysql_innodb_cluster_metadata.clusters "
+            "WHERE EXISTS "
+            "("
+            "SELECT * "
+            "FROM information_schema.schemata "
+            "WHERE schema_name = 'mysql_innodb_cluster_metadata'"
+            ")",
```
sinclert-canonical, line 1731: This query is the same as [the one](https://github.com/canonical/mysql-operator/blob/5a82aa4e888b04f573c1f23f663aa9a70b0a07ee/lib/charms/mysql/v0/mysql.py#L1696-L1702) a few lines above. See comment.

tests/unit/test_mysql.py
```diff
@@ -1144,10 +1144,43 @@ def test_cluster_metadata_exists(self, _run_mysqlsh_script):
...
+
+        query = (
+            "SELECT cluster_name "
+            "FROM mysql_innodb_cluster_metadata.clusters "
+            "WHERE EXISTS "
+            "("
+            "SELECT * "
+            "FROM information_schema.schemata "
+            "WHERE schema_name = 'mysql_innodb_cluster_metadata'"
+            ")",
```
sinclert-canonical, line 1164: This is yet another occurrence of the same query. I would probably make it a constant somewhere.

src/charm.py
```diff
@@ -768,14 +784,19 @@ def _can_start(self, event: StartEvent) -> bool:
...
+                with attempt:
+                    if self.unit_initialized(raise_exceptions=True):
+                        logger.debug(
+                            "Delegate status update for start handler on initialized unit."
+                        )
+                        self._on_update_status(None)
+                        return False
+        except RetryError:
+            event.defer()
+            return False
```
sinclert-canonical, line 799: I like that, after the refactor of the `unit_initialized` method, this retry-block looks the same as most of the others within the codebase 🧡

src/charm.py
```diff
@@ -440,21 +440,32 @@ def _handle_non_online_instance_status(self, state) -> None:
...
 
             if all_states == {"offline"} and self.unit.is_leader():
+                loopback_entry_exists = self.hostname_resolution.update_etc_hosts(None)
+                if loopback_entry_exists and not snap_service_operation(
+                    CHARMED_MYSQL_SNAP_NAME, CHARMED_MYSQLD_SERVICE, "restart"
+                ):
+                    self.unit.status = BlockedStatus(
+                        "Unable to restart mysqld before rebooting from complete outage"
+                    )
+                    return False
```
sinclert-canonical, line 450: Why is this needed? I could not find any reference to it on the PR description.

sinclert-canonical (commented):
Really like the depth of the PR description! Left a few comments.

src/hostname_resolution.py
sinclert-canonical: What's the team stand on the walrus operator? 😬

sinclert-canonical: Unsure

sinclert-canonical: https://tenor.com/en-GB/view/unsure-hmm-well-im-not-sure-gif-8356182

sinclert-canonical: https://tenor.com/en-GB/view/unsure-hmm-well-im-not-sure-gif-8356182

(joking, I never relied on it, but it is valid syntax).

src/charm.py
shayancanonical: In environments provisioned by MAAS, an entry is added to `/etc/hosts` by MAAS which resolves the hostname to a loopback address. Using a loopback address (`127.0.1.1`) as the `bind-address` in mysql causes an exception by mysqld

`hostname_resolution.update_etc_hosts` filters out such loopback entries and adds an entry resolving the hostname to the IP address of the machine. once the entries in `/etc/hosts` are updated, we need to restart mysqld because the `bind-address` in mysqld's config is set to the FQDN/hostname of the machine (since the IP address of the machine can change, and if this changes, it can cause issues with group replication)

sinclert-canonical: Thanks for the detailed explanation.

tests/unit/test_mysql.py
shayancanonical: will consolidate, but i actually like keeping the hardcoded unittest queries separate from the declaration of the queries in the mysql charm lib to ensure that anyone changing the query is intentionally doing so (as they'll need to update the unittest)

shayancanonical: consolidated in 62c01c24aed1fc888bcc6cff7f2b8e09e5834bbc

lib/charms/mysql/v0/mysql.py
shayancanonical: extracted in 62c01c24aed1fc888bcc6cff7f2b8e09e5834bbc

src/hostname_resolution.py
shayancanonical: use the walrus in 62c01c24aed1fc888bcc6cff7f2b8e09e5834bbc

lib/charms/mysql/v0/mysql.py
```diff
@@ -1674,35 +1689,63 @@ def is_cluster_in_cluster_set(self, cluster_name: str) -> Optional[bool]:
 
         return cluster_name in cs_status["clusters"]
 
-    def cluster_metadata_exists(self, from_instance: str) -> bool:
+    def cluster_metadata_exists(self, from_instance: Optional[str] = None) -> bool:
```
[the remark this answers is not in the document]
shayancanonical: added in 62c01c24aed1fc888bcc6cff7f2b8e09e5834bbc

lib/charms/mysql/v0/mysql.py
```diff
@@ -1674,35 +1689,63 @@ def is_cluster_in_cluster_set(self, cluster_name: str) -> Optional[bool]:
...
+                )
+            except MySQLClientError:
+                logger.warning(f"Failed to check if cluster metadata exists {from_instance=}")
+                raise MySQLClusterMetadataExistsError(
+                    f"Failed to check if cluster metadata exists {from_instance=}"
+                )
+
+            return self.cluster_name in output
+
+        check_cluster_metadata_query = (
```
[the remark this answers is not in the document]
shayancanonical: done in 62c01c24aed1fc888bcc6cff7f2b8e09e5834bbc

taurus-forever (approved):
LGTM, the final decision is on @paulomach here.

src/charm.py
```diff
@@ -440,21 +440,32 @@ def _handle_non_online_instance_status(self, state) -> None:
...
                     logger.error("Failed to reboot cluster from complete outage.")
                     self.unit.status = BlockedStatus("failed to recover cluster.")
-                finally:
-                    return
+                    return False
 
             if self._mysql.is_cluster_auto_rejoin_ongoing():
                 logger.info("Cluster auto-rejoin attempts are still ongoing.")
-            else:
+            elif all_states != {"offline"}:
```
sinclert-canonical, line 468: Please double-check that this is needed, as discussed privately.

sinclert-canonical (approved):

paulomach (approved):
Let's go. It will be a good addition to next promotion

PR #616, opened 2025-02-14
[DPE-6653] Signal outdated charm libs with a label

sinclert-canonical:
This PR moves the charm libs check to its own workflow, signalling outdated charm libs with a label, instead of making CI fail (Similar to what PostgreSQL repositories do).


a-velasco:
> is this an `enhancement` given that it's not user facing & only affects pr ci? cc @a-velasco

This PR should be marked as `not bug or enhancement`. 

Since these labels are used only for release notes automations, `enhancement` and `bug` should only be used for changes that affect the user of the charm, not developers.

paulomach (approved):

PR #619, opened 2025-02-19
[DPE-6666] Remove async-replication test workaround

sinclert-canonical:
This PR removes a work-around present in the async-replication tests, to circumvent a bug that old versions of MySQL had, where under continuous write conditions, a _follower_ cluster could not scale up its replicas.

Even if the official [MySQL bug](http://bugs.mysql.com/114624) is not marked as solved, the issue is no longer reproducible on MySQL 8.0.41.

### DIFF
--- a/tests/integration/high_availability/test_async_replication.py
+++ b/tests/integration/high_availability/test_async_replication.py
@@ -300,14 +300,6 @@ async def test_remove_relation_and_relate(
     first_model: Model, second_model: Model, continuous_writes
 ) -> None:
     """Test removing and re-relating the two mysql clusters."""
-    logger.info("Stopping continuous writes after 5s")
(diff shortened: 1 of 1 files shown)


paulomach (commented):
@sinclert-canonical have you tested also putting the `test_deploy_router_and_app` before `test_create_replication` ? 

The bugfix should also work for this case, i.e. having load on initial setup of async replication

sinclert-canonical:
> Have you tested also putting the `test_deploy_router_and_app` before `test_create_replication` ?

No, I thought the work around did not involve test ordering. Will re-order those two tests and see whether CI passes.

paulomach (approved):
Noice! Now we are fully testing it

shayancanonical (approved):

PR #625, opened 2025-02-25
[DPE-1531] Port changes from K8s PR to avoid --initialize-insecure

shayancanonical:
## Issue
There are changes in https://github.com/canonical/mysql-k8s-operator/pull/596/files which need to be ported over
Noticed that we are writing `ALTER USER` SQL queries before we change its permissions (leading to the password to be exposed if the chown command fails)

## Solution
Port changes
Chown the file first before writing queries to it


paulomach (approved):
I am unsure about the need for changing `chown`, since the file will be deleted by the context manager. The only thing this is doing is reducing the time the password will be in the file, or there's something else?

taurus-forever (approved):

PR #627, opened 2025-02-27
Avoid using incompatible juju debug-log options in test helper

shayancanonical:
## Issue
```
$ juju debug-log --model=test-upgrade-rollback-incompat-pz5a --level=WARNING --lines=100 --no-tail
ERROR setting --no-tail and --lines not valid
```

## Solution
Avoid using `juju debug-log --lines --no-tail`. Replace `--lines` with `--limit`

### DIFF
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -328,7 +328,7 @@ async def get_model_logs(ops_test: OpsTest, log_level: str, log_lines: int = 100
         "debug-log",
         f"--model={ops_test.model.info.name}",
         f"--level={log_level}",
-        f"--lines={log_lines}",
+        f"--limit={log_lines}",
         "--no-tail",
     )
 


paulomach (approved):

sinclert-canonical:
After a private conversation with Shayan it was made clear that these changes are due to `juju debug-log` config flag compatibility changes introduced in Juju 3.6.3, the helper method was working fine before.

> Sinclert: I would assume replacing --lines by --limit would render --no-tail irrelevant. Am I missing something?

> Shayan: I just confirmed that with juju 2.9.51 and juju 2.9.52, the --no-tail is necessary

We also clarified that the usage of `--no-tail`, although irrelevant for Juju 3.6.X, it is needed for Juju 2.9.X environments.

sinclert-canonical (approved):

PR #650, opened 2025-07-01
[DPE-7648] Fix access to not populated instance label

sinclert-canonical:
This PR aims to fix an intermittent [error](https://github.com/canonical/mysql-k8s-operator/actions/runs/15968330351/job/45048983537#step:12:715) Paulo Machado and I saw when a multiple-unit cluster is starting.

The symptom is that a specific unit calling the get_cluster_endpoints method (`mysql-k8s-1` from the error trace) is unable to find the key corresponding to another unit (`mysql-k8s-2` from the error trace), within the dictionary of Juju unit labels. The root cause seems to be the time lapse between a MySQL replica instance starting to replicate transactions, and the time where it is fully caught up, time where the custom label gets applied.

image

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -133,7 +133,7 @@ LIBID = "8c1428f06b1b4ec8bf98b7d980a38a8c"
 # Increment this major API version when introducing breaking changes
 LIBAPI = 0
 
-LIBPATCH = 88
+LIBPATCH = 89
 
 UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
 UNIT_ADD_LOCKNAME = "unit-add"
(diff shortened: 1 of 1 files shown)


paulomach (approved):
That will do. Thanks for a better solution

paulomach:
@sinclert-canonical , on merge, replicate to k8s

sinclert-canonical:
> @sinclert-canonical , on merge, replicate to k8s

Will do. Current approach does not seem to completely fix the issue, as you probably have seen in the MySQL K8s [PR](https://github.com/canonical/mysql-k8s-operator/pull/629).

- The 2nd run failed with `KeyError: 'mysql-k8s-2'` .
- The 3rd run failed with `KeyError: 'mysql-k8s-0.mysql-k8s-endpoints.testing.svc.cluster.local.:3306'`

Given the disparity of the KeyErrors, I think the problem lies somewhere else... I will try to make [VM event handler](https://github.com/canonical/mysql-operator/blob/411fb45ecfe200d22b46af9cbacaf72e8eb09da7/src/relations/mysql_provider.py#L58-L79) and [K8s event handler](https://github.com/canonical/mysql-k8s-operator/blob/bda0b6b7b5b53fcec95c0cb16dfc04534d3d0766/src/relations/mysql_provider.py#L213-L231) as similar as possible. They seem to be counterparts of each other, **despite not performing the same safety-checks**.

taurus-forever (approved):
Nice catch!

Issue #654, opened 2025-07-14
Endpoints set to ":3306"


carlcsaposs-canonical:
## Steps to reproduce

mysql-router-operator scheduled tests (https://github.com/canonical/mysql-router-operator/actions?query=event%3Aschedule)
Not 100% reproducible, also appears to be resolved after some time (but not before causing error on router)

## Expected behavior
If `endpoints` is set in the databag for the server <-> router relation, a host is always present

## Actual behavior

Endpoints is set to `:3306` without a host

## Versions

Operating system: Ubuntu 24.04

Juju CLI: 3.6.8

Juju agent: 3.6.8

Charm revision: 381

LXD: 5.21.3-c5ae129

## Log output

Juju debug log: 

[logs-integration-test-test_subordinate_charms.py-juju36_ubuntu22-arm64.zip](https://github.com/user-attachments/files/21212034/logs-integration-test-test_subordinate_charms.py-juju36_ubuntu22-arm64.zip)

```
unit-mysql-router-0: 08:15:53 INFO unit.mysql-router/0.juju-log backend-database:8: database created at 2025-07-14 08:15:53.326525
unit-mysql-router-0: 08:15:53 DEBUG unit.mysql-router/0.juju-log backend-database:8: Emitting custom event <DatabaseCreatedEvent via MachineSubordinateRouterCharm/DatabaseRequires[backend-database]/on/database_created[88]>.
unit-mysql-router-0: 08:15:53 WARNING unit.mysql-router/0.juju-log backend-database:8: FOOBAR: :3306
unit-mysql-router-0: 08:15:53 DEBUG unit.mysql-router/0.juju-log backend-database:8: self._app_workload_container_version='138' self._unit_workload_container_versions={'mysql-router/0': '138'}
unit-mysql-router-0: 08:15:53 DEBUG unit.mysql-router/0.juju-log backend-database:8: self._app_workload_container_version='138' self._unit_workload_container_versions={'mysql-router/0': '138'}
unit-mysql-router-0: 08:15:53 DEBUG unit.mysql-router/0.juju-log backend-database:8: Setting self._current_versions={'charm': '1+d40d6bb-dirty', 'workload': '8.0.41'} in upgrade peer relation app databag
unit-mysql-router-0: 08:15:53 DEBUG unit.mysql-router/0.juju-log backend-database:8: Set self._current_versions={'charm': '1+d40d6bb-dirty', 'workload': '8.0.41'} in upgrade peer relation app databag
[...]

PR #655, opened 2025-07-14
[MISC] Fix Juju Spaces lack of addresses

sinclert-canonical:
This PR fixes issue https://github.com/canonical/mysql-operator/issues/654 by updating the relation endpoints within the MySQL provider `_update_endpoints` [method](https://github.com/canonical/mysql-operator/blob/cb6707bc9ddef57629dea517155fd46bf0dab05f/src/relations/mysql_provider.py#L153-L178). This way, we ensure that the databag fields are set before accessing them in the `get_cluster_endpoints` method.

This approach is an alternative to moving the Juju Spaces addresses update to the `config_changed` hook, as that has proven challenging given current unit tests configuration.

### DIFF
--- a/src/relations/mysql_provider.py
+++ b/src/relations/mysql_provider.py
@@ -157,6 +157,8 @@ class MySQLProvider(Object):
             relation_id (int): The id of the relation
             remote_app (str): The name of the remote application
         """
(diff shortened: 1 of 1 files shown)


paulomach (approved):
read, understood and agreed

PR #658, opened 2025-07-24
[DPE-7798] Update MySQL to 8.0.42 & add s390x arch

carlcsaposs-canonical:
Since there are only 3 IS-hosted s390x github runners, will add s390x integration tests in a separate PR to avoid blocking PRs


sinclert-canonical (changes requested):
I have one requirement and one suggestion for this PR:

- Requirement: we would need to update the [architecture.py](https://github.com/canonical/mysql-operator/blob/main/lib/charms/mysql/v0/architecture.py) charm lib to also account for the `s390x` architecture. Once done, we can define a new integration test in [test_architecture.py](https://github.com/canonical/mysql-operator/blob/main/tests/integration/test_architecture.py), only to run on `s390x` runners (therefore, defining a new integration test marker).
- Suggestion: As far as I know, comments such as [this one](https://github.com/canonical/mysql-operator/blob/ac405ecadd6e079d538d66977c982888685b8eeb/tests/integration/high_availability/test_upgrade_rollback_incompat.py#L34-L35) were meant for whenever a MySQL 8.0.35+ version introduced a breaking change, so we should not update those comments yet. However, most of the tests in the [test_upgrade_rollback_incompat.py](https://github.com/canonical/mysql-operator/blob/main/tests/integration/high_availability/test_upgrade_rollback_incompat.py) file mentioned how we need to _Support ARM_. I think we need to update those to also mention `s390x`.

paulomach (approved):

Issue #649, opened 2025-06-30
On action `get_cluster_status` failure, results are set as if success 


paulomach:
## Steps to reproduce

1. on an offline unit, call `get_cluster_status` action

## Expected behavior

Event fail

## Actual behavior

On [line](https://github.com/canonical/mysql-operator/blob/411fb45ecfe200d22b46af9cbacaf72e8eb09da7/lib/charms/mysql/v0/mysql.py#L580) the failure to get status will set a result instead of calling `event.fail` , rendering a non standard output for the user.

## Versions

All

PR #663, opened 2025-08-04
bug : replace event.set_results(success=False) with event.fail() in get-cluster-status

arjun11-malik:
## Issue
Fixes Issue #649 
## Solution

Replaced the improper usage of `event.set_results(success=False)` with `event.fail("...")`.
Also added a unit test for `get-cluster-status` action to cover both success and failure scenarios.

## Checklist
- [ ] I have added or updated any relevant documentation.
- [ ] I have cleaned any remaining cloud resources from my accounts.

### DIFF
--- a/tests/unit/test_action_get_cluster_status.py
+++ b/tests/unit/test_action_get_cluster_status.py
@@ -61,10 +61,13 @@ def test_get_cluster_status_action_success(harness):
         evt.fail.assert_not_called()
 
 
-@pytest.mark.parametrize("backend_result", [
-    {"error": RuntimeError("boom")},
-    {"response": None},  # silent failure
-])
+@pytest.mark.parametrize(
+    "backend_result",
+    [
+        {"error": RuntimeError("boom")},
+        {"response": None},  # silent failure
+    ],
+)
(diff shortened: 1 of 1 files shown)


lib/charms/mysql/v0/mysql.py
```diff
@@ -571,23 +571,26 @@ def _on_set_password(self, event: ActionEvent) -> None:
...
+        try:
+            if event.params.get("cluster-set"):
+                logger.debug("Getting cluster set status")
+                status = self._mysql.get_cluster_set_status(extended=0)
+            else:
+                logger.debug("Getting cluster status")
+                status = self._mysql.get_cluster_status()
+
+            if not status:
+                event.fail("Failed to read cluster status. See logs for more information.")
```
paulomach, line 583: @arjun11-malik make sure that the output change is not being misinterpreted on the the failing tests

arjun11-malik: I’ve double-checked every integration test that invokes get-cluster-status—they only wait for the action to complete and then read out results["status"]. None of them inspect the old success=False payload or rely on its exact shape, so switching to event.fail() on empty status won’t change their behavior.

PR #671, opened 2025-08-14
fix: [DPE-7404] quorum loss recovery and test fixes

paulomach:
## Issue

- Test for quorum loss is not really provoking a quorum loss.
- Corrected test expose other issues with the implementation of the promote action and rejoin procedures.
- charm automated GR instance recovery goes boom when units try to rejoin at the same time.

## Solution

- Fix test to ensure two (out of three) instances are killed at the same time  (instead of one after the other), which will cause quorum loss.
- Fix call to shell api command that recovers from quorum loss.
- Serialize the recovery of the GR instances to avoid GR hang.
- Tweaked status setting to better display quorum loss.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -507,13 +507,14 @@ class MySQLOperatorCharm(MySQLCharmBase, TypedCharmBase[CharmConfig]):
             logger.warning("Instance does not have ONLINE peers. Cannot perform manual rejoin")
             return
 
-        if self._mysql.are_locks_acquired(from_instance=cluster_primary):
-            logger.info("waiting: cluster lock is held")
-            return
         # add random delay to mitigate collisions when multiple units are rejoining
         # due the difference between the time we test for locks and acquire them
         # Not used for cryptographic purpose
         sleep(random.uniform(0, 1.5))  # noqa: S311
+
+        if self._mysql.are_locks_acquired(from_instance=cluster_primary):
+            logger.info("waiting: cluster lock is held")
+            return
         try:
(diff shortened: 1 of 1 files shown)


tests/integration/high_availability/test_primary_switchover.py
```diff
@@ -56,13 +58,30 @@ def test_cluster_failover_after_majority_loss(juju: Juju, highly_available_clust
...
-    for unit in [non_primary_units.pop(), primary_unit]:
-        machine_name = get_unit_machine(juju, app_name, unit)
-        run(["lxc", "restart", machine_name], check=True)
+    logging.info("Kill all but one unit to simulate majority loss...")
+    units_to_kill = [non_primary_units.pop(), primary_unit]
+    machine_name = []
+    for unit in units_to_kill:
+        machine_name.append(get_unit_machine(juju, app_name, unit))
+
+    run(["lxc", "restart", "--force", machine_name[0], machine_name[1]], check=True)
```
dragomirp, line 67: Shouldn't this rather be `stop`? Wouldn't the restarted units eventually go live and rejoin?

paulomach: Because we force stop 2 (out of 3) simultaneously, there will be no quorum for reelecting a new primary since only one remained. At the time the 2 restart ones come back up, they can rejoin if there's no primary. And that's exactly the scenario we need to simulate here.
One could also just recover the remaining unit, and start others after, but we test more doing the restart.

dragomirp (approved):

PR #675, opened 2025-08-19
docs: [DPE-7404] Promote-to-primary docs

paulomach:
Add/update scoped (unit/cluster) `promote-to-primary` action.


paulomach:
@a-velasco I need help with the troubleshooting index part of this PR. Currently the index is what we have as doc. Should I split it?

PR #679, opened 2025-09-04
[DPE-8050] Backup action cluster checks

sinclert-canonical:
This PR adds a cluster health check to the `create-backup` action, which can be overridden using the `force` param.

The MySQL cluster could be in any of the states listed in this documentation [page](https://dev.mysql.com/doc/mysql-shell/8.0/en/monitoring-innodb-cluster.html#:~:text=status%3A%20The%20status%20of%20the%20InnoDB%20Cluster.). From those, only `OK` and `OK_PARTIAL` are considered safe, given that the rest of the states could introduce downtime (like the no-tolerance states) or data-loss (lost quorum, error, etc), when performing a backup.

### Additional changes
- Unnecessary mocking using `PropertyMock` class has been simplified.

### DIFF
--- a/docs/how-to/back-up-and-restore/create-a-backup.md
+++ b/docs/how-to/back-up-and-restore/create-a-backup.md
@@ -15,9 +15,18 @@ Once `juju status` shows Charmed MySQL as `active` and `idle`, you can create yo
 juju run mysql/leader create-backup
 ```
 
-If you have a cluster of one unit, you can run the `create-backup` action on `mysql-k8s/leader` (which will also be the primary unit). 
+If you have a cluster of one unit, you can run the `create-backup` action on the leader (which will also be the primary unit).
+Otherwise, you must run the `create-backup` action on a non-primary unit. To find the primary, see `juju status` or
+run `juju run mysql/leader get-cluster-status` to find the primary unit.
 
-Otherwise, you must run the `create-backup` action on a non-primary unit. To find the primary, see `juju status` or run `juju run mysql-k8s/leader get-cluster-status` to find the primary unit).
+The `create-backup` action validates that the unit in charge of the backup is healthy, by:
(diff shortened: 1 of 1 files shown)


lib/charms/mysql/v0/backups.py
```diff
@@ -355,6 +364,25 @@ def _on_create_backup(self, event: ActionEvent) -> None:
         })
         self.charm._on_update_status(None)
 
+    def _can_cluster_perform_backup(self, force: bool) -> tuple[bool, str | None]:
+        """Validates whether this cluster can perform a backup.
+
+        Returns: tuple of (success, error_message)
+        """
+        cluster_status = self.charm._mysql.get_cluster_status()
```
paulomach, line 372: Force should take precedence (or be assessed out of the function).
The point is, if we force we bypass all health checking.

And maybe we should rename the parameter to something like `skip_health_check` for clear intention?

actions.yaml
```diff
@@ -45,6 +45,12 @@ set-tls-private-key:
 create-backup:
   description: Create a database backup using xtrabackup.
     S3 credentials are retrieved from a relation with the S3 integrator charm.
+  params:
+    force:
```
paulomach, line 49: maybe change to `skip_checks` or something. Not sure

sinclert-canonical:
PR branch rebased from `main` branch.

actions.yaml
```diff
@@ -45,6 +45,12 @@ set-tls-private-key:
...
+      type: boolean
+      default: False
+      description: |
+        Whether to ignore cluster health concerns, and create the backup regardless.
```
paulomach, line 53:
```suggestion
        Whether to ignore cluster health concerns, and create the backup regardless.
        Use it with caution, as it can potentially create a backup from stale data.
```

paulomach (commented):
lgtm, but please include extra text in action and update documentation.

actions.yaml
sinclert-canonical: Done in commit https://github.com/canonical/mysql-operator/pull/679/commits/aee86163dcce913045663aac60ace1920b5d7fcb.

paulomach (approved):

PR #684, opened 2025-09-16
[DPE-7322] Tweak database DBA role naming

sinclert-canonical:
This PR updates the database-level DBA role naming in order to lift the restriction of having database names limited to 20 characters (down from the 32 character limit that MySQL supports). The agreed approach is to append a two-digit number at the end of the DBA role name, in order to distinguish it from other DBA role whose first set of characters is the same. 

### Proposed approach

Paulo and I think that using the Juju relation ID is a bad idea. It will heavily trim the space dedicated to the, potentially pruned database name (as relation IDs are several digits long), and introduced consistency issues when relations get broken, and new relations targeting the same database are created.

Instead, we propose the simple approach of incrementing an implicit counter, dictated by the number of roles which the naming schema would collide with.

### DIFF
--- a/lib/charms/mysql/v0/mysql.py
+++ b/lib/charms/mysql/v0/mysql.py
@@ -127,7 +127,7 @@ LIBID = "8c1428f06b1b4ec8bf98b7d980a38a8c"
 # Increment this major API version when introducing breaking changes
 LIBAPI = 0
 
-LIBPATCH = 94
+LIBPATCH = 95
 
 UNIT_TEARDOWN_LOCKNAME = "unit-teardown"
 UNIT_ADD_LOCKNAME = "unit-add"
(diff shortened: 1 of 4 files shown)


lib/charms/mysql/v0/mysql.py
```diff
@@ -1534,6 +1545,7 @@ def create_database(self, database: str) -> None:
         )
 
         try:
+            logger.info(f"Creating application database {database} and DBA role {role_name}")
```
paulomach, line 1548: pedantic nit
```suggestion
            logger.info(f"Creating application {database=} and DBA {role_name=}")
```

paulomach (approved):

lib/charms/mysql/v0/mysql.py
sinclert-canonical: Done in commit https://github.com/canonical/mysql-operator/pull/684/commits/1c44568a6304c15cae614428f209611d5f51b796.

paulomach (approved):

PR #714, opened 2025-11-20
[DPE-8923] Migrate tests/integration/integration/{relations,roles} to Jubilant

astrojuanlu:
Mirrors https://github.com/canonical/mysql-k8s-operator/pull/683

## Checklist
- [ ] I have added or updated any relevant documentation.
- [ ] I have cleaned any remaining cloud resources from my accounts.

### DIFF
--- a/tests/integration/integration/relations/test_database_new.py
+++ b/tests/integration/integration/relations/test_database_new.py
@@ -14,7 +14,6 @@ from utils import generate_random_password
 from ... import markers
 from ...helpers import execute_queries_on_unit, get_read_only_endpoint_ips
 from ...helpers_ha import (
-    CHARM_METADATA,
     MINUTE_SECS,
     get_app_units,
     get_mysql_primary_unit,
@@ -28,7 +27,7 @@ from ...helpers_ha import (
 
 logger = logging.getLogger(__name__)
 
-DATABASE_APP_NAME = CHARM_METADATA["name"]
+DATABASE_APP_NAME = "mysql"
 CLUSTER_NAME = "test_cluster"
 
 APPLICATION_APP_NAME = "mysql-test-app"
--- a/tests/integration/integration/relations/test_shared_db_new.py
+++ b/tests/integration/integration/relations/test_shared_db_new.py
@@ -9,7 +9,6 @@ import pytest
 from jubilant_backports import Juju
 
 from ...helpers_ha import (
-    CHARM_METADATA,
     check_keystone_users_existence,
     check_successful_keystone_migration,
     get_app_units,
@@ -24,7 +23,7 @@ from ...helpers_ha import (
 
 logger = logging.getLogger(__name__)
 
-APP_NAME = CHARM_METADATA["name"]
+APP_NAME = "mysql"
 CLUSTER_NAME = "test_cluster"
 KEYSTONE_APP_NAME = "keystone"
 ANOTHER_KEYSTONE_APP_NAME = "another-keystone"
--- a/tests/integration/integration/relations/test_db_router_new.py
+++ b/tests/integration/integration/relations/test_db_router_new.py
@@ -9,7 +9,6 @@ import pytest
 from jubilant_backports import Juju
 
 from ...helpers_ha import (
-    CHARM_METADATA,
     check_keystone_users_existence,
     check_successful_keystone_migration,
     get_app_units,
@@ -23,7 +22,7 @@ from ...helpers_ha import (
 
 logger = logging.getLogger(__name__)
 
-APP_NAME = CHARM_METADATA["name"]
+APP_NAME = "mysql"
 CLUSTER_NAME = "test_cluster"
 KEYSTONE_APP_NAME = "keystone"
 KEYSTONE_MYSQLROUTER_APP_NAME = "keystone-mysql-router"
--- a/tests/integration/integration/relations/test_relation_mysql_legacy_new.py
+++ b/tests/integration/integration/relations/test_relation_mysql_legacy_new.py
@@ -12,14 +12,13 @@ from ...helpers import (
     is_connection_possible,
 )
 from ...helpers_ha import (
-    CHARM_METADATA,
     get_legacy_mysql_credentials,
     wait_for_apps_status,
 )
 
 logger = logging.getLogger(__name__)
 
-DATABASE_APP_NAME = CHARM_METADATA["name"]
+DATABASE_APP_NAME = "mysql"
 CLUSTER_NAME = "test_cluster"
 
 APPLICATION_APP_NAME = "mysql-test-app"
--- a/tests/integration/integration/roles/test_database_dba_role_new.py
+++ b/tests/integration/integration/roles/test_database_dba_role_new.py
@@ -11,7 +11,6 @@ from mysql.connector.errors import ProgrammingError
 
 from ...helpers import execute_queries_on_unit
 from ...helpers_ha import (
-    CHARM_METADATA,
     MINUTE_SECS,
     get_app_units,
     get_mysql_primary_unit,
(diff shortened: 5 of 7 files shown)


paulomach (approved):
lgtm - not sure what's going on with arm runners though

tests/integration/helpers_ha.py
```diff
@@ -29,6 +35,43 @@
...
+        "raise_on_warnings": False,
+        "raw": raw,
+    }
+
+    with MysqlConnector(config, commit) as cursor:
+        for query in queries:
+            cursor.execute(query)
+        output = list(itertools.chain(*cursor.fetchall()))
+
+    return output
```
sinclert-canonical: I would rather leave the refactor of the _execute-query-in-unit_ function to a different PR. Reasons:

1. **Proper refactoring is needed:** hopefully, we could have it as a class receiving username, password & IP address on initialization time; query, commit & raw when the method is called.
2. **Consistency with K8s counterpart:** the sibling repository, although having a _sync_ version of this function, it does not have it on the `helpers_ha.py`. Let's keep Python modules / code structure as close as possible between the two repos, to ease the future merging.
3. **Consistency within the repo itself:** introducing a _sync_ version of this function would mean slightly refactoring all previously migrated Jubilant test, to take advantage of the new helper. However, that is not proposed in this PR.

---

I am not saying having a _sync_ version will be unwelcome. I am asking to decouple the two efforts.

tests/integration/helpers_ha.py
```diff
@@ -220,6 +275,26 @@ def get_relation_data(juju: Juju, app_name: str, rel_name: str) -> list[dict]:
...
+    relation_data = get_relation_data(juju=juju, app_name=app_name, rel_name=relation_name)
+    read_only_endpoint_ips = get_read_only_endpoint_ips(relation_data)
+    # check that the number of read-only-endpoints is correct
+    assert len(get_app_units(juju, app_name)) - 1 == len(read_only_endpoint_ips)
+    app_ips = [
+        get_unit_ip(juju, app_name, unit_name) for unit_name in get_app_units(juju, app_name)
+    ]
+    # check that endpoints are the one of the application
+    for read_endpoint_ip in read_only_endpoint_ips:
+        assert read_endpoint_ip in app_ips
```
sinclert-canonical: When refactoring the HA integration tests I made sure that helpers that are only used in a single tests (i.e. very specific to what is being tested) lived on the same module as the test itself. Reason is: we should prevent the helpers module to become a dumpster where _everything goes_ (as we had it before).

This helper seems to only be used in a single module, let's move it there.

tests/integration/helpers_ha.py
```diff
@@ -268,25 +343,92 @@ def get_mysql_primary_unit(juju: Juju, app_name: str, unit_name: str | None = No
...
+    params = {"username": username}
+    if password is not None:
+        params["password"] = password
+
+    rotate_task = juju.run(
+        unit=unit_name,
+        action="set-password",
+        params=params,
+    )
+    rotate_task.raise_on_failure()
```
sinclert-canonical: This helper seems to only be used in a single module, let's move it there.

tests/integration/integration/relations/test_database_new.py
```diff
@@ -0,0 +1,245 @@
...
+def test_password_rotation(juju: Juju):
+    """Rotate password and confirm changes."""
+    random_unit = get_app_units(juju, DATABASE_APP_NAME)[-1]
+
+    old_credentials = get_mysql_server_credentials(juju, random_unit)
+
+    # get primary unit first, need that to invoke set-password action
+    primary_unit_name = get_mysql_primary_unit(juju, DATABASE_APP_NAME, random_unit)
+    primary_unit_address = get_unit_ip(juju, DATABASE_APP_NAME, primary_unit_name)
+    logger.debug("Primary unit detected before password rotation is %s", primary_unit_address)
```
sinclert-canonical: This comment will only be left once, but it applies to **all occurrences** of these lines.

I think we can simplify the fetching of a random unit by executing the password rotation in the primary itself. To do so, just leave the last argument of the `get_mysql_primary_unit` function empty, so the helper will use the _leader_ to get who the primary is.

tests/integration/integration/relations/test_shared_db_new.py
```diff
@@ -0,0 +1,162 @@
...
+    )
+
+    juju.wait(
+        ready=lambda status: (
+            (status.apps[keystone_application_name].app_status.current == "waiting")
+            and all(
+                unit.workload_status.current == "blocked"
+                for unit in status.get_units(keystone_application_name).values()
+            )
+        ),
```
sinclert-canonical: Same comment as before.

tests/integration/helpers_ha.py
astrojuanlu: That's fair enough, will do

tests/integration/helpers_ha.py
sinclert-canonical: I would say a ~1400 LOC PR is long enough to avoid doing more refactors than those strictly necessary, but alas, this is not a blocker for me.

Just bear in mind what we discussed privately: let's try to keep MySQL VM + MySQL K8s **as similar as possible** to ease the merging for future us.

tests/integration/integration/relations/test_database_new.py
```diff
@@ -0,0 +1,276 @@
...
+        relation_name: The name of the relation
+    """
+    relation_data = get_relation_data(juju=juju, app_name=app_name, rel_name=relation_name)
+    read_only_endpoint_ips = get_read_only_endpoint_ips(relation_data)
+    # check that the number of read-only-endpoints is correct
+    if len(get_app_units(juju, app_name)) - 1 != len(read_only_endpoint_ips):
+        return False
+    app_ips = [
+        get_unit_ip(juju, app_name, unit_name) for unit_name in get_app_units(juju, app_name)
+    ]
```
sinclert-canonical: Should this me called `unit_ips`?

tests/integration/integration/relations/test_database_new.py
sinclert-canonical: In MySQL K8s I always placed the module-level helpers at the end of the file.

Let's try to be consistent and move them there.

sinclert-canonical (commented):
Provided 2nd round of review.

tests/integration/helpers_ha.py
```diff
@@ -126,11 +180,12 @@ def scale_app_units(juju: Juju, app_name: str, num_units: int) -> None:
...
-        ready=wait_for_apps_status(jubilant_backports.all_active, app_name),
-        error=jubilant_backports.any_blocked,
-        timeout=20 * MINUTE_SECS,
-    )
+
+    if num_units > 0:
+        juju.wait(
+            ready=wait_for_apps_status(jubilant_backports.all_active, app_name),
+            timeout=20 * MINUTE_SECS,
+        )
```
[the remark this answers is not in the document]
sinclert-canonical: Got it! Thanks for checking.

tests/integration/conftest.py
```diff
@@ -9,6 +10,8 @@
 
 from . import architecture
 
+logging.getLogger("jubilant.wait").setLevel(logging.WARNING)
```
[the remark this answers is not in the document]
sinclert-canonical: If we know that setting the Jubilant log level in the `conftest.py` works (it seems so by the CI jobs output), we should simplify every single Python module that sets it up individually.

Similarly to [this comment](https://github.com/canonical/mysql-operator/pull/714#discussion_r2570746521): I think we should either refactor it completely from previously migrated Jubilant tests, or preserve the style we have been following so far, and leave this simplification for a follow-up effort.

PR #727, opened 2025-12-09
[DPE-8916] Use mysql-shell-client package

sinclert-canonical:
This PR removes a whole _horizontal slice_ of the charm code, to rely on [mysql-shell-client](https://github.com/canonical/mysql-shell-client) package.

Exceptions to this extraction are:
- The creation of the MySQL Router roles (see [code](https://github.com/canonical/mysql-operator/blob/5b7d312dc3d5b7d0733c7cb097ea70db2f32b69e/lib/charms/mysql/v0/mysql.py#L1258-L1267)), as providing admin privileges to such roles is a bad pattern.
- The creation of the MySQL Router user (see [code](https://github.com/canonical/mysql-operator/blob/5b7d312dc3d5b7d0733c7cb097ea70db2f32b69e/lib/charms/mysql/v0/mysql.py#L1410-L1418)), as its only consumer will be removed in MySQL 8.4.
- The configuration of the charm system users (see [code](https://github.com/canonical/mysql-operator/blob/5b7d312dc3d5b7d0733c7cb097ea70db2f32b69e/lib/charms/mysql/v0/mysql.py#L1309-L1325)), as it is extremely specific to how we do things.

---

### Highlights:
- All hard-coded MySQL instance roles have been replaced by constants (when not comparing to existing data-bags).
- All hard-coded MySQL instance states have been replaced by constants (when not comparing to existing data-bags).
- Unified ways to count the number of cluster members in a given state (see previous [method 1](https://github.com/canonical/mysql-operator/blob/e451f10c525f9ad982ec6c3b2c3564ae17be64cd/lib/charms/mysql/v0/mysql.py#L2425-L2459), [method 2](https://github.com/canonical/mysql-operator/blob/e451f10c525f9ad982ec6c3b2c3564ae17be64cd/src/upgrade.py#L109-L118)).
- Unified ways to kill open connections (see previous [method 1](https://github.com/canonical/mysql-operator/blob/aecbf0ffb5011df5fd39e3b39383dcaec37f45c3/lib/charms/mysql/v0/mysql.py#L3564-L3586), [method 2](https://github.com/canonical/mysql-operator/blob/aecbf0ffb5011df5fd39e3b39383dcaec37f45c3/lib/charms/mysql/v0/mysql.py#L3588-L3610)).
- ~Removed `lock_instance` argument from `add_instance_to_cluster` charm lib function~.
- ~Removed `lock_instance` argument from `remove_instance_to_cluster` charm lib function~.

### Review strategy

Given the amount of changes proposed on this PR, I recommend reviewing commit by commit. When reaching https://github.com/canonical/mysql-operator/pull/727/commits/57a78eed3597d9e0f41abd6db13e4a02507015c0, please compared side-by-side:

- Previous `mysql` [charm lib](https://github.com/canonical/mysql-operator/blob/aecbf0ffb5011df5fd39e3b39383dcaec37f45c3/lib/charms/mysql/v0/mysql.py).
- Proposed `mysql` [charm lib](https://github.com/canonical/mysql-operator/blob/57a78eed3597d9e0f41abd6db13e4a02507015c0/lib/charms/mysql/v0/mysql.py).

### DIFF
--- a/lib/charms/mysql/v0/async_replication.py
+++ b/lib/charms/mysql/v0/async_replication.py
@@ -26,6 +26,12 @@ from constants import (
     SERVER_CONFIG_PASSWORD_KEY,
     SERVER_CONFIG_USERNAME,
 )
+from mysql_shell.models import (
+    ClusterGlobalStatus,
+    ClusterRole,
+    InstanceRole,
+    InstanceState,
+)
 from ops import (
     ActionEvent,
     ActiveStatus,
@@ -51,7 +57,7 @@ logger = logging.getLogger(__name__)
 # The unique Charmhub library identifier, never change it
 LIBID = "4de21f1a022c4e2c87ac8e672ec16f6a"
 LIBAPI = 0
-LIBPATCH = 11
+LIBPATCH = 12
 
 PYDEPS = ["mysql_shell_client ~= 0.6"]
 
@@ -98,21 +104,30 @@ class MySQLAsyncReplication(Object):
         """Current cluster set role of the unit, after the relation is established."""
         is_replica = self._charm._mysql.is_cluster_replica()
 
+        # TODO:
+        #  Remove `.lower()` when migrating to MySQL 8.4
+        #  (when breaking changes are allowed)
         if is_replica:
-            cluster_role = "replica"
+            cluster_role = ClusterRole.REPLICA.lower()
         elif is_replica is False:
-            cluster_role = "primary"
+            cluster_role = ClusterRole.PRIMARY.lower()
         else:
+            # TODO:
+            #  Uppercase when migrating to MySQL 8.4
+            #  (when breaking changes are allowed)
             cluster_role = "unset"
 
-        _, instance_role = self._charm._mysql.get_member_state()
+        instance_role = self._charm._mysql.get_member_role()
 
         if self.model.get_relation(RELATION_CONSUMER):
             relation_side = RELATION_CONSUMER
         else:
             relation_side = RELATION_OFFER
 
-        return ClusterSetInstanceState(cluster_role, instance_role, relation_side)
+        # TODO:
+        #  Remove `.lower()` when migrating to MySQL 8.4
+        #  (when breaking changes are allowed)
+        return ClusterSetInstanceState(cluster_role, instance_role.lower(), relation_side)
 
     @property
     def cluster_name(self) -> str:
@@ -180,7 +195,13 @@ class MySQLAsyncReplication(Object):
         """Handle the async relation being broken from either side."""
         # Remove the replica cluster, if this is the primary
 
-        if self.role.cluster_role in ("replica", "unset") and not self._charm.removing_unit:
+        # TODO:
+        #  Remove `.lower()` when migrating to MySQL 8.4
+        #  (when breaking changes are allowed)
+        if (
+            self.role.cluster_role in (ClusterRole.REPLICA.lower(), "unset")
+            and not self._charm.removing_unit
+        ):
             # The cluster being removed is a replica cluster
             # role is `unset` when the primary cluster dissolved the replica before
             # this hook execution i.e. was faster on running the handler
@@ -221,7 +242,10 @@ class MySQLAsyncReplication(Object):
             # set flag to persist removed from cluster-set state
             self._charm.app_peer_data["removed-from-cluster-set"] = "true"
 
-        elif self.role.cluster_role == "primary":
+        # TODO:
+        #  Remove `.lower()` when migrating to MySQL 8.4
+        #  (when breaking changes are allowed)
+        elif self.role.cluster_role == ClusterRole.PRIMARY.lower():
             if self._charm.unit.is_leader():
                 # only leader units can remove replica clusters
                 remote_data = event.relation.data.get(event.relation.app) or {}
@@ -232,8 +256,8 @@ class MySQLAsyncReplication(Object):
 
                         # force removal when cluster is invalidated
                         force = self._charm._mysql.get_replica_cluster_status(cluster_name) in [
-                            "invalidated",
-                            "unknown",
+                            ClusterGlobalStatus.INVALIDATED,
+                            ClusterGlobalStatus.UNKNOWN,
                         ]
 
                         self._charm._mysql.remove_replica_cluster(cluster_name, force=force)
@@ -272,8 +296,8 @@ class MySQLAsyncReplication(Object):
             return
 
         status = self._charm._mysql.get_replica_cluster_status(cluster)
-        if status != "invalidated":
-            message = f"Cluster {status=}. Only `invalidated` clusters can be rejoined"
+        if status != ClusterGlobalStatus.INVALIDATED:
+            message = f"Cluster {status=}. Only `INVALIDATED` clusters can be rejoined"
             event.fail(message)
             logger.info(message)
             return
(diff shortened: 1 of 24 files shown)


lib/charms/mysql/v0/async_replication.py
```diff
@@ -795,7 +822,7 @@ def _on_consumer_changed(self, event):  # noqa: C901
 
             self._charm.unit.status = MaintenanceStatus("Dissolving replica cluster")
             logger.info("Dissolving replica cluster")
-            self._charm._mysql.dissolve_cluster()
+            self._charm._mysql.dissolve_cluster(force=True)
```
paulomach: ~~why force?~~ found it. ok

lib/charms/mysql/v0/mysql.py
```diff
@@ -1068,23 +1038,65 @@ def __init__(
...
+        )
 
-        Args:
-            user: User name.
-            host: Host name, default to unit address.
-        """
-        if host and ":" in host:
-            # strip port from address
-            host = host.split(":")[0]
+    def _build_cluster_tcp_executor(self, host: str, port: int = 3306) -> BaseExecutor:
```
paulomach: Why not use `ADMIN_PORT` for cluster executor?

src/mysql_vm_helpers.py
```diff
@@ -728,149 +724,6 @@ def restart_mysql_exporter(self) -> None:
         self.stop_mysql_exporter()
         self.connect_mysql_exporter()
 
-    def _run_mysqlsh_script(
```
paulomach: removal feels good :)

paulomach (commented):
Great work, I like the direction this is going.
out of a couple comments just one is a blocker (about admin port), otherise lgtm

lib/charms/mysql/v0/async_replication.py
sinclert-canonical: Honestly we do not need the keyword argument here, as I made the default value `True`.

lib/charms/mysql/v0/async_replication.py
```diff
@@ -53,6 +53,8 @@
 LIBAPI = 0
 LIBPATCH = 11
 
+PYDEPS = ["mysql_shell_client ~= 0.6"]
```
astrojuanlu: Do you plan to keep the 0.* series backwards compatible?

lib/charms/mysql/v0/mysql.py
```diff
@@ -119,8 +136,7 @@ def __init__(
 
 # Increment this major API version when introducing breaking changes
 LIBAPI = 0
-
-LIBPATCH = 99
+LIBPATCH = 100
```
astrojuanlu: Nice milestone 😄

lib/charms/mysql/v0/mysql.py
```diff
@@ -625,6 +618,7 @@ def _on_promote_to_primary(self, event: ActionEvent) -> None:
             except MySQLForceQuorumFromInstanceError:
                 logger.exception("Failed to force quorum from instance")
                 event.fail("Failed to force quorum from instance. See logs for more information.")
+                return
```
astrojuanlu: Uh were these bugs before? I see the "success: True" line below is reached without this `return` (same with the one right below)

lib/charms/mysql/v0/mysql.py
```diff
@@ -1312,211 +1316,113 @@ def configure_mysql_system_users(self) -> None:
...
-                f"session.run_sql(\"GRANT SELECT, INSERT, UPDATE, DELETE, EXECUTE ON mysql_innodb_cluster_metadata.* TO '{username}'@'{hostname}';\")",
-                f"session.run_sql(\"GRANT SELECT ON mysql.user TO '{username}'@'{hostname}';\")",
-                f"session.run_sql(\"GRANT SELECT ON performance_schema.replication_group_members TO '{username}'@'{hostname}';\")",
-                f"session.run_sql(\"GRANT SELECT ON performance_schema.replication_group_member_stats TO '{username}'@'{hostname}';\")",
-                f"session.run_sql(\"GRANT SELECT ON performance_schema.global_variables TO '{username}'@'{hostname}';\")",
-            )
+        primary_address = self.get_cluster_primary_address()
+        primary_executor = self._build_instance_tcp_executor(primary_address)
+
+        escaped_user_attributes = json.dumps({"unit_name": unit_name}).replace('"', r"\"")
```
astrojuanlu: Isn't this what `StringQueryQuoter.escape` does? I see that quoters don't belong in executors though 🤔

astrojuanlu (dismissed):
Left quite a few questions, but otherwise I'd say let's get this merged. Amazing job @sinclert-canonical 👏🏼

lib/charms/mysql/v0/mysql.py
```diff
@@ -610,7 +603,7 @@ def _on_promote_to_primary(self, event: ActionEvent) -> None:
         if event.params.get("scope") != "unit":
             return
 
-        if self._mysql.is_unit_primary(self.unit_label):
+        if self._mysql.get_primary_label() == self.unit_label:
```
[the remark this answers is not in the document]
sinclert-canonical: I don't think there was a strong reason to do so. I believe I moved it to the `charm.py` module thinking it was highly used there, but that is not the case.

Can revert back easily.

lib/charms/mysql/v0/mysql.py
sinclert-canonical: Yes, this was an **actual bug**, which Paulo and I discovered while debugging some tests. An incorrect message (i.e. _unit is already primary_) was displayed regardless of crash root cause.

lib/charms/mysql/v0/mysql.py
```diff
@@ -2507,266 +2069,175 @@ def remove_instance(  # noqa: C901
...
+            raise MySQLGetClusterPrimaryAddressError() from e
 
-        if not matches:
-            return None
+        cluster_status = status["defaultReplicaSet"]["status"]
+        cluster_address = status["defaultReplicaSet"]["primary"]
 
-        return matches.group(1)
+        if cluster_status == ClusterStatus.NO_QUORUM:
+            raise MySQLGetClusterPrimaryAddressError()
```
sinclert-canonical, line 2199: The original method had an implicit pre-flight check where if the cluster status is considered a dangerous one to get cluster status from, fails. For the K8s operator (client to this charm lib), we need to replicate this behavior.

lib/charms/mysql/v0/mysql.py
```diff
@@ -2507,266 +2069,175 @@ def remove_instance(  # noqa: C901
...
+        primary_cluster = status["primaryCluster"]
+        primary_cluster_status = status["clusters"][primary_cluster]["globalStatus"]
+        primary_cluster_address = status["clusters"][primary_cluster]["primary"]
 
-        address = matches.group(1)
-        if ":" in address:
-            # strip port from address
-            address = address.split(":")[0]
+        if primary_cluster_status == ClusterGlobalStatus.INVALIDATED:
+            raise MySQLGetClusterPrimaryAddressError()
```
sinclert-canonical, line 2223: Given the proposed validation when fetching the a cluster primary address, I figured we could do the same with the cluster-set primary address.

astrojuanlu (dismissed):
:shipit:

src/charm.py
```diff
@@ -571,25 +577,31 @@ def _on_update_status(self, _) -> None:  # noqa: C901
...
-            self.unit_peer_data["member-role"] = role
-            self.unit_peer_data["member-state"] = state
-        except (MySQLNoMemberStateError, MySQLUnableToGetMemberStateError):
-            role = self.unit_peer_data["member-role"] = "unknown"
-            state = self.unit_peer_data["member-state"] = "unreachable"
+            role = self._mysql.get_member_role()
+            state = self._mysql.get_member_state()
+        except MySQLUnableToGetMemberStateError:
+            role = "UNKNOWN"
+            state = "UNREACHABLE"
```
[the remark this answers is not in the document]
paulomach: not mere luck, but definitely a safe-default

paulomach (dismissed):
:shipit:

paulomach (approved):

astrojuanlu (approved):

PR #735, opened 2026-01-15
Add optional flags to provides endpoints

rpbritton:
Filed by @canonical/solutions-qa 

## Issue

Since provides endpoints can be non-optional, can the default value of `optional` is [false](https://canonical-charmcraft.readthedocs-hosted.com/stable/reference/files/charmcraft-yaml-file/#endpoint-role-endpoint-name-optional), these endpoints are assumed non-optional.

## Solution

Set `optional` flag on provides endpoint to indicate correct optionality.

## Checklist
- [x] I have added or updated any relevant documentation.
- [x] I have cleaned any remaining cloud resources from my accounts.

### DIFF
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -32,8 +32,10 @@ peers:
 provides:
   database:
     interface: mysql_client
+    optional: false
   mysql:
     interface: mysql
+    optional: false
   db-router:
     interface: mysql-router
     optional: true
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -36,14 +36,18 @@ provides:
     interface: mysql
   db-router:
     interface: mysql-router
+    optional: true
   shared-db:
     interface: mysql-shared
+    optional: true
   cos-agent:
     interface: cos_agent
     limit: 1
+    optional: true
   replication-offer:
     interface: mysql_async
     limit: 1
+    optional: true
 
 requires:
   certificates:


sinclert-canonical (dismissed):

astrojuanlu (dismissed):
> Include the optional key in all endpoint definitions, rather than relying on the default value to indicate that the relation is required. Although this field is not enforced by Juju, including it makes it clear to users (and other tools) whether the relation is required.

([source](https://documentation.ubuntu.com/charmcraft/stable/reference/files/charmcraft-yaml-file/#endpoint-role-endpoint-name-optional))

LGTM! In fact, maybe we should go ahead and set `optional: false` in the other two.

rpbritton:
> > Include the optional key in all endpoint definitions, rather than relying on the default value to indicate that the relation is required. Although this field is not enforced by Juju, including it makes it clear to users (and other tools) whether the relation is required.
> 
> ([source](https://documentation.ubuntu.com/charmcraft/stable/reference/files/charmcraft-yaml-file/#endpoint-role-endpoint-name-optional))
> 
> LGTM! In fact, maybe we should go ahead and set `optional: false` in the other two.

That is a good find in the documentation. I've set that on the other two.

paulomach (approved):

astrojuanlu:
Backporting to the new repo at https://github.com/canonical/mysql-operators/pull/51, closing this PR. Thanks @rpbritton !

